PR #21, opened 2024-06-09
Remove `DynError` and some chores

al8n:


iduartgomez:
I don’t think we should be using anyhow here, but rather more structured errors using this error. We can then convert to anyhow at wherever this API is being consumed.

PR #27, opened 2025-01-27
Various changes required for river integration

sanity:
…ract loading


sanity:
@iduartgomez @netsirius It's a bit of a mish-mash of changes but I think this is ready to review however in order to make progress I had incorporated the changes from the `add-subscribe-flag` branch. I could back out those changes related to the new `subscribe` field but it would be a PITA so prefer to avoid. Thoughts?

PR #70, opened 2026-04-30
fix!: stringify NodeDiagnosticsResponse.contract_states keys

sanity:
## Problem

`NodeDiagnosticsResponse.contract_states` was `HashMap<ContractKey, ContractState>`. `ContractKey` derives `Serialize` as a struct (`{instance, code}`), but `serde_json` rejects non-string JSON object keys with `key must be a string`. The wire path between freenet-core and clients uses bincode, which does not care about key types, so the bug was invisible until the `freenet service report` binary tried to serialize the response to JSON for upload — every report from a node hosting at least one contract has been silently uploading an empty `network_status` since the field was added in June 2025.

The user-facing report bug was masked further by report.rs swallowing the WS query error until #3897/#3898 surfaced it. See freenet/freenet-core#3987 for the user-impact details and the full story.

## Solution

Switch the key type to `String` (Base58 contract id, matching the existing `ContractKey::Display` impl). This is the convention every other field in this struct already uses:
- `NodeInfo.peer_id: String`
- `NetworkInfo.connected_peers: Vec<(String, String)>`
- `ConnectedPeerInfo.peer_id: String`, `address: String`
- `ContractHostingEntry.contract_key: String`

`contract_states` was the lone outlier. The fix makes the type consistent with sibling diagnostic fields and unblocks the JSON serialization path at the source.

## Wire-format compatibility

This is a **bincode wire-format break** for `NodeDiagnosticsResponse`. Older clients built against 0.6.x that decode the `HostResponse::QueryResponse(QueryResponse::NodeDiagnostics(_))` variant will fail to deserialize the new payload.

Known consumers of `NodeDiagnosticsResponse`:
- `freenet-core/crates/fdev/src/diagnostics.rs` (the `fdev query`/`diagnostics` commands) — already calls `key.to_string()` on the map key, so it is forward-compatible source-wise. fdev is shipped with freenet-core releases and rebuilt in lockstep.
- `freenet-core/crates/core/src/bin/commands/report.rs` (the `freenet service report` binary) — was already broken on the old shape; freenet/freenet-core#3989 lands a local workaround so the report fix doesn't depend on this stdlib release.
- `freenet-core/crates/core/src/node/network_bridge/p2p_protoc.rs` (the producer side, inserts into the map) — will need to call `.to_string()` on the contract key when freenet-core bumps to a stdlib that includes this change.

No external consumers are known. Riverctl and River UI do not call `NodeDiagnostics`.

## Testing

Adds `node_diagnostics_response_tests::node_diagnostics_response_json_round_trips` in `client_events.rs`, which constructs a populated `NodeDiagnosticsResponse` and asserts:
- `serde_json::to_string(&response)` succeeds (the case the bug broke)
- The output round-trips back to JSON
- The Base58 key is preserved verbatim

Why this regression test wasn't there before: there was no JSON round-trip test for `NodeDiagnosticsResponse` in either repo. The wire path used in production is bincode, which does not exercise the JSON-key-must-be-string constraint. This test closes that gap at the source — any future field whose key type does not serialize as a string would break this test.

`cargo test --workspace` clean (19 stdlib tests + 1 new). `cargo fmt --check` clean. `cargo clippy --workspace --all-targets -- -D warnings` clean.

## Release notes

Updated `CHANGELOG.md` under `[Unreleased]`. This change is queued for the next stdlib release; it does not need to land before freenet/freenet-core#3989, which fixes the user-facing report bug independently.

## Refs

- freenet/freenet-core#3987 — user-facing bug report
- freenet/freenet-core#3989 — companion freenet-core fix that does not require this stdlib change to land

[AI-assisted - Claude]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,32 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed (wire-format break in `NodeDiagnosticsResponse`)
+- `NodeDiagnosticsResponse.contract_states` is now
+  `HashMap<String, ContractState>` (Base58 contract id) instead of
+  `HashMap<ContractKey, ContractState>`. The previous type had a
+  derived `Serialize` for `ContractKey` that emitted a struct
+  (`{instance, code}`), which `serde_json` rejects because JSON object
+  keys must be strings — every diagnostic report from a node hosting at
+  least one contract uploaded with empty `network_status`. The new key
+  matches the convention every other field in this struct already uses
+  (`peer_id: String`, `connected_peers: Vec<(String, String)>`,
+  `ContractHostingEntry::contract_key: String`). See
+  freenet/freenet-core#3987.
+- This is a bincode wire-format break for `NodeDiagnosticsResponse`.
+  Older clients (built against 0.6.x) that decode the
+  `HostResponse::QueryResponse(QueryResponse::NodeDiagnostics(_))`
+  variant will fail to deserialize the new payload. The only known
+  in-tree consumer is `fdev`'s diagnostics command, which is shipped
+  with `freenet-core` and rebuilt in lockstep. `freenet service report`
+  was already broken on the old shape and works with the new shape
+  via the local fix in freenet/freenet-core#3989.
+- Added a `serde_json` round-trip regression test for
+  `NodeDiagnosticsResponse` to prevent the same class of bug from
+  reappearing — any future field whose key type does not serialize as
+  a string would break this test at the source.
+
 ## [0.6.0] - 2026-04-13
 
 ### Changed (source-level breaking, wire-compatible)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Unreleased]
+## [0.7.0]
 
 ### Fixed (wire-format break in `NodeDiagnosticsResponse`)
 - `NodeDiagnosticsResponse.contract_states` is now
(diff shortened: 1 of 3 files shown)


sanity:
Multi-model review pass complete. Four agents (code-first, testing, skeptical, big-picture) plus Codex CLI; convergent + complementary findings addressed in 502bd42:

- **Bumped version to 0.7.0** in \`rust/Cargo.toml\` (big-picture-reviewer): the \`fix!:\` is a bincode wire-format break and the manifest should reflect that intent now rather than relying on the maintainer to remember at release time.

- **Expanded CHANGELOG consumer list** (skeptical, big-picture, code-first reviewers, all flagged this independently): added \`p2p_protoc.rs:1900,1916\` (producer side, will fail to compile on bump), \`freenet-test-network/src/network.rs:1093\` (out-of-tree map lookup keyed by ContractKey), and the \`freenet-core report.rs::diagnostics_to_json\` workaround that becomes redundant. The original CHANGELOG only listed fdev.

- **Documented bidirectional wire-format break** (Codex P1): both directions break in mixed-version setups (newer clients can't decode older nodes' payloads either, not just the reverse). Made explicit in the Compatibility section.

- **Documented \`code_hash\` drop** (skeptical-reviewer): \`ContractKey::Display\` only emits \`instance\`. No in-tree consumer reads code_hash from this map, but worth recording for future consumers.

- **Fully populated regression test** (testing-reviewer; matches the same finding addressed on the freenet-core PR): every \`Option\` is \`Some\`, every \`Vec\` is non-empty, every value distinguishable, and the test asserts exactly six top-level keys. Now any future field added to \`NodeDiagnosticsResponse\` with a non-string-keyed map gets exercised.

- **Added bincode round-trip assertion** (testing-reviewer): the new wire format is the contract; pinning bincode shape alongside JSON shape catches silent layout drift.

- **Filed freenet/freenet-core#3993** (big-picture-reviewer's recommendation): tracks the post-bump cleanup in freenet-core (stringify keys at producer sites, simplify \`diagnostics_to_json\`, delete the bug-pinning sentinel test).

Reviewer recommendations explicitly NOT addressed in this PR, with rationale:

- **Backward-compatible deserializer** (Codex P1): overkill for a debug-only payload; in-tree consumers run matched versions; the bidirectional break is now documented and accepted as part of the major bump.
- **Bincode wire-format pin tests for \`HostResponse\`/\`QueryResponse\` variants** (testing + big-picture reviewers): scope creep for this PR; better as a follow-up that pins all variants uniformly per the pattern from #66.
- **Wrap regression test in \`HostResponse::QueryResponse(...)\` for end-to-end coverage** (testing-reviewer): defense-in-depth, not blocking; the field-level test is the primary contract.

Codex's other observation that the original PR description's claim 'every diagnostic report uploaded with empty network_status' is technically about the freenet-core path that's already fixed downstream — the stdlib fix is independent — is addressed implicitly by the new CHANGELOG which explicitly notes that #3989 already shipped the workaround.

All 19 stdlib tests pass; \`cargo clippy --workspace --all-targets -- -D warnings\` clean; \`cargo fmt --check\` clean.

[AI-assisted - Claude]

PR #71, opened 2026-05-06
feat: mirror freenet-stdlib history to Freenet-hosted demo repo

sanity:
## Summary

- Adds `.github/workflows/mirror-to-freenet.yml` which calls the reusable workflow at `freenet/freenet-git/.github/workflows/mirror-repo.yml` (merged in [freenet-git#17](https://github.com/freenet/freenet-git/pull/17)).
- Triggers: push to main, workflow_dispatch, daily cron (12:53 UTC) as a safety net for missed pushes.
- Mode is `history` — the full git history is small enough (~720 KiB pack) to publish, so consumers get a real `git log` not just a HEAD snapshot.

## Demo URL

`freenet:96rknpy1GYhZ/freenet-stdlib` (replaces the old `2pyvKxrozxgT` demo, which was created from a passphrase-encrypted bundle that doesn't fit a CI workflow).

## Required secrets (already configured on this repo)

- `FREENET_GIT_IDENTITY_BUNDLE_BASE64` — base64 of the no-passphrase bundle that owns the new prefix.
- `FREENET_GIT_WS_URL` — same value as `RIVER_GATEWAY_URL` on freenet-core.

## Test plan

- [x] Reusable workflow merged on the freenet-git side ([#17](https://github.com/freenet/freenet-git/pull/17)).
- [x] Initial state for `96rknpy1GYhZ/freenet-stdlib` already published manually (clonable today via `git clone freenet::96rknpy1GYhZ/freenet-stdlib`).
- [x] Rescue workflow on the freenet-git side smoke-tested — green in 3m45s against the live demo URLs.
- [ ] After merge: confirm a push to main triggers the workflow and produces an updated state.

[AI-assisted - Claude]


sanity:
Addressed Codex P1 + Skeptical H1/H2: pinned `@main` to the merge SHA of freenet-git#17 (`362e547c7035d28968651d2f4106fbfa891c31e2`) and replaced `secrets: inherit` with explicit pass-through of just the two secrets the reusable workflow declares as required.

**Pushed back / deferring:**
- **Skeptical H3** (snapshot orphan commit picks up wall-clock time → daily cron creates a new commit hash even when the source is unchanged). Real issue, but the fix belongs in the reusable workflow in freenet-git, not these caller PRs. Will follow up with a separate PR there to set `GIT_AUTHOR_DATE`/`GIT_COMMITTER_DATE` from the source commit's `%cI` so the orphan commit is deterministic.
- **Skeptical M2** (no `paths:` filter — every commit to main mirrors). Reasonable optimization but not blocking; would tune `paths-ignore: ['*.md', 'docs/**', '.github/**']` after observing real workflow usage.
- **Skeptical M1, L1, L3**: comment polish / minor; not blocking. Caller `workflow_dispatch:` is a usable manual recovery button per the verified-safe analysis.

[AI-assisted - Claude]

sanity:
Ran the missing internal reviewers (code-first, testing, big-picture) at the user's request -- I had only run skeptical + Codex earlier and judged it sufficient for a 27-line workflow file, which was the wrong call to make without checking first.

Bottom line from all four agents: **no blockers**. Reviewers' new findings were:

**Code-first should-fix (real comment-accuracy issues — fixed in 95eab5a):**
- Cron offset comment claimed gateway-contention protection that doesn't exist (the reusable workflow's `concurrency` group keys on `freenet_repo`, so different prefixes are in different groups). Reworded to "staggered for log readability only."
- SHA-pin comment implied the pin closed the whole supply chain, but `cargo install freenet-git --locked` still pulls `latest` from crates.io. Added a note acknowledging the gap and the conditions under which it'd be worth tightening (override `freenet_git_version`).

**Testing reviewer:** wanted a `workflow_dispatch` smoke run from the PR branch before merge. Tried; `gh workflow run --ref` requires the workflow file to exist on the default branch first. Accepting the residual risk: the underlying primitives are validated (rescue workflow smoke-tested green; manual push of initial state for both demo URLs verified end-to-end). The only unknown is GitHub's cross-repo `uses:` plumbing, which is well-trodden.

**Big-picture:** "no blockers, approve and merge." Notes are nits (SHA-bump procedure not documented, weekly-vs-daily cron, fork guard). Not addressing in this PR.

**Pushed back / deferred:**
- Code-first nit #5 (branch fallback if main renamed): not applicable for this repo.
- Code-first nit #8 (fork guard `if: github.repository == ...`): defensible but adds noise; freenet-stdlib forks would fail loudly at "Write identity bundle" step which is acceptable.
- Code-first nit #9 (SHA-linked URL in comment): cosmetic.
- Code-first nit #10 (deprecation note for old `2pyvKxrozxgT` URL): worth a separate doc PR, not blocking.
- Big-picture #2 (weekly cron): worth observing actual runs first.

Also accidentally bundled three regenerated flatbuffer files in commit `a29f0e7` (worktree-local flatc version mismatch); reverted in `95eab5a`. PR diff is now back to one file, 46 lines added. CI re-running.

[AI-assisted - Claude]

PR #82, opened 2026-07-02
feat: scheduled wakeup primitive for delegates (ScheduleWakeup / WakeupFired)

sanity:
## Problem

Delegates have no way to schedule future execution, so dApps needing periodic background work (key rotation, TTL pruning, scheduled publication) must push it into a UI/client sync loop, which stops the moment the tab closes.

Driving use case: River private-rooms weekly secret rotation (freenet/river#228).

## Approach — and the shape changed during review

```rust
// host function, freenet_delegate_management namespace
DelegateCtx::schedule_wakeup(after: Duration, tag: &[u8]) -> Result<(), i64>

// inbound wire variant, tag 9
InboundDelegateMsg::WakeupFired { tag: Vec<u8> }
```

Plus `MAX_WAKEUP_TAG_BYTES` (128) and `MIN_WAKEUP_DELAY` (1s).

An earlier draft made the outbound half a wire variant too. **That would have silently eaten users' messages**, which is the finding that reshaped this PR.

### Why the outbound half is a host function

A delegate's outbound messages are serialized as **one batch** — a single `bincode::serialize` of `Result<Vec<OutboundDelegateMsg>, DelegateError>` at `delegate_interface.rs` — and decoded whole by the host. An outbound variant the host does not recognize fails the **entire batch**.

So a delegate built against this release, returning `[ApplicationMessage(reply), ScheduleWakeup{..}]` to a current-release node, would have **the reply discarded along with the wakeup**. The user's action silently does nothing, with no error anywhere.

That is not an edge case: it is the direction ordinary rollout produces **every time**, because stdlib ships before core by policy.

A host function fails the opposite way — an unimported function fails at **instantiation**, with a named missing import. Loud, once, at load time, before the delegate has touched anything.

| | old delegate → new host | new delegate → old host |
|---|---|---|
| outbound wire variant | fine | **whole batch dropped**, silently, per message |
| host function | unaffected if not imported | fails to **instantiate**, named import, at load |

freenet-stdlib#128 makes exactly this argument for itself. Shipping this as an outbound variant would have put two opposite answers to one question in the same release.

**`WakeupFired` stays an inbound wire variant** because that direction has no equivalent hazard: a delegate that cannot schedule never receives one. `OUTBOUND_VARIANT_COUNT` therefore returns to 9; inbound stays at 10.

## `after` is a delay, not a deadline

The **host** measures it, from when it receives the call.

A delay needs no clock on the delegate side, and forecloses nothing: if absolute scheduling is ever wanted it is `target - now` expressed with this same argument. A delegate re-arming inside its own `WakeupFired` handler simply asks for the same delay again, so the recurring case needs neither a clock nor a timestamp on the fire — which avoids adding a field the host could only fill meaningfully or leave permanently useless.

It also removes a failure mode rather than an ambiguity: wall-clock steps (NTP, a manual change) become the host scheduler's business rather than a semantic left undefined, and there is no value that fails to encode.

**`MIN_WAKEUP_DELAY` (1s)**: delays below it are clamped **up**. A delegate re-arming with a zero delay would otherwise spin the node in a tight wake loop — moving from a deadline to a delay removed the "deadline in the past" spelling of that hazard, not the hazard. The floor is a host behaviour; the constant states the contract so an author can rely on it. The guarantee is "not before"; nothing promises precision.

**`MAX_WAKEUP_TAG_BYTES` (128)**, enforced before the host call. The host bounds how many wakeups a delegate may hold pending, but a count cap alone does not bound memory while the per-item size is caller-chosen — the same defect as a cache bounded by entry count holding caller-controlled values.

## `WakeupFired` carries no `DelegateContext`

Two independent reasons, and the second was found by checking core rather than reasoning about it.

**It is not a reply.** A context is per-conversation working state the host hands back on a reply. A wakeup opens a conversation; `tag` supplies the correlation. Carrying one would commit the host to persisting delegate context across arbitrary wall-clock time — a storage subsystem, not a field, and #5467 Phase 3.

**The cache could not supply a coherent value anyway.** freenet-core's `DelegateContextCache` is keyed **per delegate**, not per conversation, and prunes after `DELEGATE_CONTEXT_TTL` (10 minutes, `native_api.rs`). Any wakeup worth scheduling outlives that, so the context present at scheduling time is gone. And if a live context *does* exist inside that window, it belongs to a different in-flight exchange — reading it during a wakeup would be reading another conversation's state. Core's own doc says state needing to survive longer belongs in the delegate's secrets.

The exemption is **named** in the tests rather than wildcarded, and asserts `get_context()` is `None`, so if the variant ever grows a context the exemption fails rather than silently widening.

## Durability is a requirement on the host, and is **not implemented**

A week-long delay is only useful if pending wakeups survive a node restart. **No host does this today**, and nothing in this crate can make it true — freenet-core#3972 must.

Stated as an obligation rather than a guarantee because the precedent runs the wrong way: `DELEGATE_SUBSCRIPTIONS`, the one comparable piece of per-delegate host state, is an in-memory `LazyLock<DashMap>` (`native_api.rs:41-43`) that a restart discards entirely. An earlier draft asserted persistence as a property of this change; a wire doc claiming a durability guarantee nothing provides is worse than silence, because the next reader builds on it.

Assuming the host implements it, delivery is **at-most-once**: a wakeup is removed when it fires, so a crash between firing and the delegate finishing `process()` can lose it. A delegate needing a hard guarantee should also re-assert from startup logic.

## Compatibility

- **`WakeupFired` is appended at tag 9**, so no existing tag moves.
- **New host → old delegate**: an unknown inbound tag is a **hard bincode decode error**, not a skipped message (`invalid value: integer 9, expected variant index 0 <= i < 9`). It is safe here only because the host sends `WakeupFired` solely to a delegate that scheduled one, and a delegate that cannot import `schedule_wakeup` cannot have scheduled. **That is a requirement on freenet-core#3972, not a property of the wire format.**
- The host function is additive in both directions; see the table above.

See [WIRE-FORMAT.md](https://github.com/freenet/freenet-stdlib/pull/129) for the general rules this follows.

## Testing

`inbound_wakeup_fired_wire_format_is_stable` pins `WakeupFired`'s tag and full byte layout. Four new guards cover the host-function bounds.

**The guards were made to fail before they were trusted:**

| Mutation | Result |
|---|---|
| Tag cap check removed | `an_oversized_tag_is_refused_without_calling_the_host` **fails** ✓ |
| Off-by-one: `>` → `>=` | `a_tag_exactly_at_the_cap_is_not_refused_for_being_too_big` **fails** ✓ |

The off-WASM stub returns `Err(ERR_NOT_IN_PROCESS)`, never `Ok(())`, and that is asserted — a stub reporting success would let a host-side test read "wakeup scheduled" out of a call that scheduled nothing.

This PR was also what first exercised #98's variant-tag pin against a real second change: rebasing onto it produced four `E0004` errors and a failing unknown-tag probe, forcing the move from tag 8 to 9. A colliding append surfaced as a compile error, never as a silent renumbering.

## Verification

- `cargo test --features contract,net,testing,trace` — **130 passed, 0 failed**
- `cargo clippy --all-targets --features contract,net,testing,trace -- -D warnings` — clean
- `cargo build --target wasm32-unknown-unknown --features contract` — clean
- `cargo fmt --check` — clean

Version is **0.10.0**, unchanged in this diff — `main` already carries it from #98.

## Downstream

freenet-core#3972 must supply: the delay measured from receipt, the `MIN_WAKEUP_DELAY` clamp, a bound on pending wakeups per delegate, persistence across restart, and the opt-in property that keeps `WakeupFired` safe.

Refs: freenet/freenet-core#3972, freenet/river#228

[AI-assisted - Claude]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,32 @@
 # Changelog
 
+## [0.8.3]
+
+### Added
+- Scheduled-wakeup primitive for delegates (freenet/freenet-core#3972):
+  - `OutboundDelegateMsg::ScheduleWakeup { at: SystemTime, tag: Vec<u8> }` —
+    a delegate asks the host to deliver a wakeup at an absolute time. Opaque
+    `tag` is echoed back on fire. Re-scheduling with the same `tag` replaces
+    the prior pending wakeup for that `(delegate, tag)` pair (cancel-by-tag).
+  - `InboundDelegateMsg::WakeupFired { tag: Vec<u8> }` — delivered by the host
+    when a scheduled wakeup fires.
+  - Lets always-on delegates run periodic background work (key rotation, TTL
+    pruning, scheduled publication) without a connected UI, instead of pushing
+    that work into a client sync loop.
+
+### Compatibility
+- **Wire-compatible, source-additive.** Both variants are appended at the end
+  of their enums (bincode variant tag 8), so no existing variant's tag shifts —
+  delegate WASM compiled against an older stdlib keeps deserializing every
+  message it already understood. New wire-format pin tests
+  (`inbound_wakeup_fired_wire_format_is_stable`,
+  `outbound_schedule_wakeup_wire_format_is_stable`) freeze the tags.
+- Adding `ScheduleWakeup` is a **source-level break for exhaustive `match`
+  sites on `OutboundDelegateMsg`** (which is intentionally *not*
+  `#[non_exhaustive]`, so the host is forced to handle every outbound variant).
+  `InboundDelegateMsg` remains `#[non_exhaustive]`; unknown inbound variants are
+  forwarded to the delegate WASM unchanged.
+
 ## [0.8.0]
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -243,14 +243,26 @@ reply to something an older delegate cannot have sent.
 
 The host measures it from when it receives the call. A delay needs no clock on
 the delegate side, and if absolute scheduling is ever wanted it is `target - now`
-in terms of this same argument — so nothing is foreclosed. Delays below
-`MIN_WAKEUP_DELAY` are clamped up to it, because a delegate re-arming inside its
-own handler with a zero delay would otherwise spin the node in a tight wake loop.
-The guarantee is "not before"; nothing promises precision.
+in terms of this same argument — so nothing is foreclosed. The guarantee is
+"not before"; nothing promises precision.
 
-`tag` is capped at `MAX_WAKEUP_TAG_BYTES`, enforced before the host call. The
-host bounds how many wakeups a delegate may hold pending, but a count cap alone
-does not bound memory while the per-item size is caller-chosen.
+`schedule_wakeup` **clamps `after` up to `MIN_WAKEUP_DELAY` itself**, so
+`Duration::ZERO` is a one-second delay rather than a tight wake loop — a delegate
+re-arming inside its own handler would otherwise spin the node.
+
+`tag` is capped at `MAX_WAKEUP_TAG_BYTES`, applied before the host call.
+
+**Both of those are fail-fast convenience for well-behaved callers, not bounds.**
+`__frnt__delegate__schedule_wakeup` is an ordinary WASM import: a delegate can
+declare its own `extern "C"` block and pass a 10 MB tag, or a zero delay, without
+ever constructing a `DelegateCtx`. The guest controls its own imports, so no
+guest-side check can bound what the host receives. **Only the host can enforce
+either, and no host does yet** — freenet-core#3972. The same is true of the
+per-delegate cap on how many wakeups may be pending, which is an obligation on
+#3972 and not something that exists today.
+
+That last point matters for reading the tag cap's rationale: a size cap is needed
+*in addition to* a count cap, and neither is in place yet.
 
 #### `WakeupFired` carries no context, and the cache explains why
 
@@ -270,63 +282,28 @@ because the precedent runs the other way — `DELEGATE_SUBSCRIPTIONS`, the one
 comparable piece of per-delegate host state, is in-memory and a restart discards
 it entirely.
 
-### Why a delay and not a deadline
-
-An earlier draft used `at: SystemTime`. **A delegate cannot fill that field.**
-Delegates compile to `wasm32-unknown-unknown`, where `SystemTime::now()` compiles
-and then panics at runtime, and freenet-core registers no temporal host function
-in any of the four delegate namespaces — so there is no value the delegate could
-compute for an absolute deadline, including the "one week from now" that motivates
-this primitive.
-
-A delay is also strictly more capable. If a clock host function is ever added,
-absolute scheduling becomes `target - now`, expressed with this same field;
-an absolute field would gain nothing it did not already need that clock for. And
-a delegate re-arming inside its `WakeupFired` handler simply asks for the same
-delay again, so the recurring case needs neither a clock nor a timestamp on the
-fire.
-
-It removes two failure modes rather than merely an ambiguity: a pre-epoch
-`SystemTime` fails to serialize on the *sender*, and wall-clock steps (NTP, a
-manual change) are now the host scheduler's business rather than a semantic left
-undefined on the wire. `Duration` encodes identically to `SystemTime` — u64
-seconds LE + u32 nanos — so the wire size is unchanged, and the pin test asserts
-the whole byte layout rather than round-tripping through this crate's own encoder.
-
-##### Durability is a requirement on the host, and is unimplemented
-
-A week-long delay is only useful if pending wakeups survive a node restart.
-**No host does this today.** freenet-core#3972 must implement it; nothing in this
-crate can make it true. Said as an obligation rather than a guarantee because the
-precedent runs the other way — `DELEGATE_SUBSCRIPTIONS`, the one comparable piece
-of per-delegate host state, is an in-memory `LazyLock<DashMap>` that a restart
-discards entirely.
-
 #### Compatibility
 
-Both variants are **appended at tag 9**, behind the unsubscribe pair at tag 8,
-so no existing variant's bincode tag moves and
-every message a deployed delegate already understood still decodes. The two
-directions are not symmetric, and the asymmetry is the part that matters:
+`WakeupFired` is **appended at tag 9**, behind the unsubscribe pair at tag 8, so
+no existing variant's bincode tag moves and every message a deployed delegate
+already understood still decodes. `inbound_wakeup_fired_wire_format_is_stable`
+pins its tag and full byte layout.
 
-- **Old delegate -> new host** (outbound): safe. An older delegate never emits
-  `ScheduleWakeup`, and every tag it does emit is unchanged.
-- **New host -> old delegate** (inbound): `WakeupFired` at a tag an older
-  delegate does not know is a **hard bincode decode error**, not a skipped
-  message. It is safe here only because the host sends `WakeupFired` solely to a
-  delegate that asked for it by emitting `ScheduleWakeup` — a delegate built
-  against an older stdlib cannot have asked, so it can never be sent one. That
-  opt-in-by-construction property is a **requirement on the host
-  implementation**, not a property of the wire format, and freenet-core#3972
-  must preserve it.
-- Adding `ScheduleWakeup` is a **source-level break for exhaustive `match`
-  sites on `OutboundDelegateMsg`**, which is intentionally *not*
-  `#[non_exhaustive]` so the host is forced to handle every outbound variant.
-  `InboundDelegateMsg` stays `#[non_exhaustive]`.
+The two directions are not symmetric:
 
-New wire-format pin tests (`inbound_wakeup_fired_wire_format_is_stable`,
-`outbound_schedule_wakeup_wire_format_is_stable`) freeze both tags.
-
+- **Old delegate -> new host**: safe, and now safe by construction rather than
+  by convention — the scheduling half is a host import, so an older delegate
+  simply does not import it. There is no outbound wire variant to mis-decode,
+  which is the whole point of the conversion above.
+- **New host -> old delegate**: `WakeupFired` at a tag an older delegate does not
+  know is a **hard bincode decode error**, not a skipped message. It is safe only
+  because the host sends it solely to a delegate that scheduled one, and a
+  delegate that cannot import `schedule_wakeup` cannot have scheduled. That
+  opt-in-by-construction property is a **requirement on freenet-core#3972**, not
+  a property of the wire format.
+- The host function itself is additive both ways: a delegate that does not import
+  it is unaffected, and one that does fails to **instantiate** on too old a node
+  with a named missing-import error.
 
 ### TypeScript SDK 0.4.0 — Breaking (npm package `@freenetorg/freenet-stdlib`)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -204,6 +204,79 @@ half is freenet-core#5565.
   only reason a field could be appended to it without corrupting anything after
   it. That property is now pinned rather than assumed.
 
+### Added — scheduled-wakeup primitive for delegates
+
+Lets an always-on delegate run periodic background work (key rotation, TTL
+pruning, scheduled publication) without a connected UI, instead of pushing that
+work into a client sync loop that stops when the tab closes.
+Driving use case: freenet/river#228. Host half: freenet/freenet-core#3972.
+
+- `OutboundDelegateMsg::ScheduleWakeup { after: Duration, tag: Vec<u8> }` — the
+  delegate asks the host to deliver a wakeup once `after` has elapsed, measured
+  by the **host** from when it receives the message. `tag` is opaque to the host
+  and echoed back on fire. Re-scheduling with the same `tag` replaces the prior
+  pending wakeup for that `(delegate, tag)` pair.
+- `InboundDelegateMsg::WakeupFired { tag: Vec<u8> }` — delivered when a
+  scheduled wakeup fires.
+
+##### Why a delay and not a deadline
+
+An earlier draft used `at: SystemTime`. **A delegate cannot fill that field.**
+Delegates compile to `wasm32-unknown-unknown`, where `SystemTime::now()` compiles
+and then panics at runtime, and freenet-core registers no temporal host function
+in any of the four delegate namespaces — so there is no value the delegate could
+compute for an absolute deadline, including the "one week from now" that motivates
+this primitive.
+
+A delay is also strictly more capable. If a clock host function is ever added,
+absolute scheduling becomes `target - now`, expressed with this same field;
+an absolute field would gain nothing it did not already need that clock for. And
+a delegate re-arming inside its `WakeupFired` handler simply asks for the same
+delay again, so the recurring case needs neither a clock nor a timestamp on the
+fire.
+
+It removes two failure modes rather than merely an ambiguity: a pre-epoch
+`SystemTime` fails to serialize on the *sender*, and wall-clock steps (NTP, a
+manual change) are now the host scheduler's business rather than a semantic left
+undefined on the wire. `Duration` encodes identically to `SystemTime` — u64
+seconds LE + u32 nanos — so the wire size is unchanged, and the pin test asserts
+the whole byte layout rather than round-tripping through this crate's own encoder.
+
+##### Durability is a requirement on the host, and is unimplemented
+
+A week-long delay is only useful if pending wakeups survive a node restart.
+**No host does this today.** freenet-core#3972 must implement it; nothing in this
+crate can make it true. Said as an obligation rather than a guarantee because the
+precedent runs the other way — `DELEGATE_SUBSCRIPTIONS`, the one comparable piece
+of per-delegate host state, is an in-memory `LazyLock<DashMap>` that a restart
+discards entirely.
+
+#### Compatibility
+
+Both variants are **appended at tag 9**, behind the unsubscribe pair at tag 8,
+so no existing variant's bincode tag moves and
+every message a deployed delegate already understood still decodes. The two
+directions are not symmetric, and the asymmetry is the part that matters:
+
+- **Old delegate -> new host** (outbound): safe. An older delegate never emits
+  `ScheduleWakeup`, and every tag it does emit is unchanged.
+- **New host -> old delegate** (inbound): `WakeupFired` at a tag an older
+  delegate does not know is a **hard bincode decode error**, not a skipped
+  message. It is safe here only because the host sends `WakeupFired` solely to a
+  delegate that asked for it by emitting `ScheduleWakeup` — a delegate built
+  against an older stdlib cannot have asked, so it can never be sent one. That
+  opt-in-by-construction property is a **requirement on the host
+  implementation**, not a property of the wire format, and freenet-core#3972
+  must preserve it.
+- Adding `ScheduleWakeup` is a **source-level break for exhaustive `match`
+  sites on `OutboundDelegateMsg`**, which is intentionally *not*
+  `#[non_exhaustive]` so the host is forced to handle every outbound variant.
+  `InboundDelegateMsg` stays `#[non_exhaustive]`.
+
+New wire-format pin tests (`inbound_wakeup_fired_wire_format_is_stable`,
+`outbound_schedule_wakeup_wire_format_is_stable`) freeze both tags.
+
+
 ### TypeScript SDK 0.4.0 — Breaking (npm package `@freenetorg/freenet-stdlib`)
 
 The npm package is versioned separately from the Rust crate. This release
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -211,15 +211,66 @@ pruning, scheduled publication) without a connected UI, instead of pushing that
 work into a client sync loop that stops when the tab closes.
 Driving use case: freenet/river#228. Host half: freenet/freenet-core#3972.
 
-- `OutboundDelegateMsg::ScheduleWakeup { after: Duration, tag: Vec<u8> }` — the
-  delegate asks the host to deliver a wakeup once `after` has elapsed, measured
-  by the **host** from when it receives the message. `tag` is opaque to the host
-  and echoed back on fire. Re-scheduling with the same `tag` replaces the prior
-  pending wakeup for that `(delegate, tag)` pair.
-- `InboundDelegateMsg::WakeupFired { tag: Vec<u8> }` — delivered when a
-  scheduled wakeup fires.
+- `DelegateCtx::schedule_wakeup(after: Duration, tag: &[u8]) -> Result<(), i64>`
+  — a **host function**, in the `freenet_delegate_management` namespace.
+- `InboundDelegateMsg::WakeupFired { tag: Vec<u8> }` — an inbound wire variant
+  at **tag 9**, delivered when a scheduled wakeup fires.
+- `MAX_WAKEUP_TAG_BYTES` (128) and `MIN_WAKEUP_DELAY` (1s).
 
-##### Why a delay and not a deadline
+#### The outbound half is a host function, not a wire variant
+
+This is the asymmetry that matters, and it is deliberate.
+
+A delegate's outbound messages are serialized as **one batch** and decoded whole
+by the host. An outbound variant the host does not recognize therefore fails the
+**entire batch** — so a delegate built against this release, returning
+`[ApplicationMessage(reply), ScheduleWakeup{..}]` to a current-release node,
+would have **the reply discarded along with the wakeup**, and the user's action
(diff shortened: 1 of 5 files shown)


sanity:
Host-side consumer: freenet/freenet-core#4666 (issue #3972). This stdlib PR must merge + publish 0.8.3 first (stdlib-first), then the core PR can build.

[AI-assisted - Claude]

sanity:
Coordination note for whoever picks this up: **tag 8 is now taken, and this PR needs to move to 9.**

Delegate unsubscribe (`UnsubscribeContractRequest` / `UnsubscribeContractResponse`, closing the half of #2830 that was specified but never built) is landing in #98 and appends to these same two enums. Both changes wanted the next free tag. @sanity settled the ordering: unsubscribe takes 8, this PR rebases onto it at 9.

Why it needed settling rather than being left to merge order: these enums are bincode-encoded with the variant index in declaration order, so two independent "append at the end" changes silently renumber each other. Nothing caught that before — #98 adds a pin asserting the tag of **every** variant in both enums, backed by an exhaustive match inside the crate so a new variant is a compile error until it is pinned. Once #98 merges, a colliding append fails CI loudly instead of shipping a wire break.

Two other things worth knowing before rebasing, neither of them new to this PR but both easy to miss:

- This branch currently **reverts four fixes merged to main since July**, so the rebase is real work rather than a fast-forward.
- It is versioned for a release slot that no longer exists. The PR says it bumps 0.8.2 to 0.8.3; crates.io is at 0.8.5, and `main` is at an unpublished 0.9.0 which already carries a breaking change (the #91 `RegisterDelegateWithPredecessors` removal). #98 is going out as part of that 0.9.0 release.

Nothing here is a judgment on the wakeup design itself, which is unchanged and still wanted — freenet-core#5467 Phase 2 tracks it, and notes that the host half (freenet-core#4666, branch `feat/3972-delegate-wakeup` @ `9139913c`) was written and Full-tier reviewed before being closed unmerged for want of a live consumer.

[AI-assisted - Claude]

sanity:
Recording a tag decision here so it does not live only in a working session, since whoever rebases this will not have seen it.

**Ian ruled that unsubscribe takes tag 8; this PR moves to tag 9.**

freenet-stdlib#98 appends `UnsubscribeContractRequest` to `OutboundDelegateMsg` and `UnsubscribeContractResponse` to `InboundDelegateMsg`, both at **tag 8**. So `ScheduleWakeup` and `WakeupFired` become **tag 9** in their respective enums, and this PR's two wire-format pin tests need their expected tags updated from 8 to 9.

Both enums are bincode-encoded with the variant index in declaration order, so two independent "append at the end" branches silently renumber each other. That is now a loud failure rather than a silent one: #98 adds a per-variant tag pin whose map is an exhaustive `match` inside the defining crate, so appending a variant without pinning it is a **compile error**, and a probe asserts the next tag along does not decode so the variant counts cannot drift. Rebasing onto it will tell you immediately if a tag collides.

Two other things worth knowing before picking this up, neither a criticism of the change itself:

1. **This branch currently reverts four fixes merged to `main` since July**, including the `fixed_size_field` decode-panic hardening and the `unknown_union_discriminant` change in `delegate_interface.rs`. That is branch staleness rather than intent, but a naive rebase would undo them.
2. **The version target has moved on twice.** The description says it bumps 0.8.2 to 0.8.3; 0.8.5 is published and `main` is already at an unreleased 0.9.0 carrying a breaking change (#91).

[AI-assisted - Claude]

sanity:
## Rebased onto current `main` (`029a31a`); now `MERGEABLE`

Was `CONFLICTING` and 15 commits stale. **Rebased, not merged** — that matters here.

### What the conflicts were, and how they were resolved

Only **two** files conflicted, and both were metadata. The two code files — `rust/src/delegate_interface.rs` and `rust/src/client_api/client_events.rs` — **auto-merged with no conflict**.

1. **`rust/Cargo.toml`** — `HEAD` had `0.9.0`, the branch had `0.8.3`. Took neither: set to **`0.10.0`** (reasoning below).
2. **`CHANGELOG.md`** — the branch added a `## [0.8.3]` section where `main` now has `## [Unreleased]`. Dropped the `[0.8.3]` header and folded the content into `[Unreleased]` as a named subsection. No text from `main` was discarded.

No behavioural judgement call was made in either resolution.

### On the "#82 reverts four fixes merged since July" warning in #98

That warning was correct about a **merge** and does not apply to a **rebase**. The branch was a single commit on top of `75e3a46`; rebasing replays only its own diff, so the four fixes on `main` — including the `fixed_size_field` decode-panic hardening and the `unknown_union_discriminant` change — are untouched. Verified: the four conflicting-file diff is `+141/-3` against `main` and touches nothing else.

### Why `0.10.0`

`0.8.3` was taken by the memory-leak fix (`8b53702`), and the branch has since been overtaken twice more. **`0.9.0` is already published** (crates.io, 2026-09-04) and `main` sits at it, so the version had to move regardless.

Appending `ScheduleWakeup` to `OutboundDelegateMsg`, which is deliberately *not* `#[non_exhaustive]`, is a **source-breaking** change for every downstream exhaustive `match`. On a 0.x line that is a minor bump, so `0.10.0` rather than `0.9.1`.

(Worth flagging separately: freenet-core still depends on `freenet-stdlib = "0.8.5"`, so the core bump for this is two releases behind, not one.)

### Verification on the rebased head `2864681`

- `cargo test --features contract,net,testing,trace` — **99 passed, 0 failed**; doc-tests 1 passed
- `cargo fmt --check` — clean

Both wire pins were confirmed to **execute**, not merely to be present — a filtered `cargo test` matching nothing prints `0 passed` and exits 0:

```
test ..::inbound_wakeup_fired_wire_format_is_stable ... ok
test ..::outbound_schedule_wakeup_wire_format_is_stable ... ok
test result: ok. 2 passed; 0 failed; ... 97 filtered out
```

### One correction to this PR's compatibility claim

The PR body says the change is "wire-compatible ... delegate WASM compiled against an older stdlib keeps deserializing everything it already understood." True, and **not sufficient** — it describes only the safe direction.

The other direction: `WakeupFired` at tag 8 sent to a delegate built against an older stdlib is a **hard bincode decode error**, not a skipped message. (Confirmed empirically against bincode 1.3.3: `invalid value: integer 8, expected variant index 0 <= i < 8`.)

That is safe here **only because the host sends `WakeupFired` solely to a delegate that asked for it by emitting `ScheduleWakeup`** — an older delegate cannot have asked, so it can never be sent one. That opt-in-by-construction property is a **requirement on freenet-core#3972**, not a property of the wire format, and nothing in the format enforces it. The CHANGELOG now says so explicitly rather than leaving it as an unstated assumption.

### Tag 8 vs 9 — still an open sequencing decision

This branch pins tag **8** on both enums, which is correct against `main` as it stands. #98 also claims tag 8 and proposes this PR move to 9.

Deliberately **not** pre-emptively renumbered to 9: if #98 does not land, a move to 9 leaves a hole at tag 8, and a hole is worse than a collision — it is a tag that decodes to nothing. Whichever of the two lands second should rebase and renumber. #98's variant-tag pin catches the collision as a **compile error**, which is that pin working as designed rather than a risk.

[AI-assisted - Claude]

sanity:
## Restacked on #98 and moved to wire tag 9

Per the sequencing decision: **#98 goes first, this rebases onto it.** #98 owns the tag pin, so it is the PR that should not be renumbered; this one moves at no cost.

**Base changed** from `main` to `feat/delegate-unsubscribe-introspection`, so the diff above is now this PR's own change only (3 files) rather than #98's plus its own. **HEAD `2bca2c0`, MERGEABLE, CI green.**

### The retag, and how it was caught

`ScheduleWakeup` / `WakeupFired` moved from tag 8 to **tag 9**, behind #98's unsubscribe pair at 8.

This was not found by reading. Rebasing produced **four `E0004` non-exhaustive-pattern errors** — #98 removed the `_ => None` wildcards from `get_context`/`get_mut_context` and added exhaustive tag maps, so the new variants could not compile without being placed:

```
error[E0004]: `InboundDelegateMsg::WakeupFired { .. }` not covered
  --> rust/src/delegate_interface.rs:611   <- get_context
  --> rust/src/delegate_interface.rs:648   <- get_mut_context
  --> rust/src/delegate_interface.rs:1618  <- pinned_inbound_tag
  --> rust/src/delegate_interface.rs:1634  <- pinned_outbound_tag
```

Then, with the tags placed, `an_unpinned_variant_fails_this_test` failed — the count constants were still 9, so tag 9 had become decodable while the probe asserted it must not be. Bumped both to 10.

That is #98's pin doing precisely what it was built to do: a colliding append surfaced as a compile error and a failing probe, not as a silent renumbering. Worth stating plainly since it is the first time that machinery has been exercised by a real second change.

### A design question this surfaced, which reviewers should confirm rather than inherit

#98 made `get_context`/`get_mut_context` exhaustive and left a comment asserting **"every variant carries a context"**. `WakeupFired` does not, so that invariant is now false, and its test `every_inbound_variant_with_a_context_exposes_it` failed.

I did **not** paper over this with a wildcard. `WakeupFired` is named as an explicit exemption — matching the pattern #98's outbound test already uses for `RequestUserInput` and `ContextUpdated` — and the exemption asserts `get_context()` is `None`, so if the variant ever grows a context the exemption fails rather than silently widening.

The reasoning for having no context, which is the part worth a second opinion:

- **`WakeupFired`**: a wakeup is an independent trigger, not the reply to a request made earlier in the same batch, so there is no context to thread. The host has none to supply either — the field could only ever be `default()`, advertising state that is always absent.
- **`ScheduleWakeup`**: fire-and-forget. Its `processed()` is `true` and it produces no response in the same batch, so there is no round trip for a context to survive.

I believe that is right, but it is a wire-shape decision and it is cheap to change now and expensive later. If a reviewer thinks a wakeup should carry context, say so before this merges.

### Compatibility, restated for tag 9

Unchanged in substance from the note above, and the asymmetry still matters:

- **Old delegate -> new host**: safe. An older delegate never emits `ScheduleWakeup`.
- **New host -> old delegate**: `WakeupFired` at tag 9 is a **hard bincode decode error** on a delegate that does not know it, not a skipped message. It is safe only because the host sends it solely to a delegate that asked by emitting `ScheduleWakeup`. **That is a requirement on freenet-core#3972, not a property of the wire format**, and nothing here enforces it.

### Verification on `2bca2c0`

- `cargo test --features contract,net,testing,trace` — **127 passed, 0 failed**
- `cargo clippy --target wasm32-unknown-unknown -- -D warnings` — clean · `cargo fmt --check` — clean
- Both retagged pins confirmed to **execute**, not merely exist:

```
inbound_wakeup_fired_wire_format_is_stable ... ok
outbound_schedule_wakeup_wire_format_is_stable ... ok
delegate_msg_variant_tags_are_pinned ... ok
an_unpinned_variant_fails_this_test ... ok
```

Still a draft, and still not for merging ahead of #98.

[AI-assisted - Claude]

sanity:
## Heads-up: this PR now has **no CI**, and that is a consequence of stacking it

`.github/workflows/ci.yml:7-8` triggers only on PRs targeting `main`:

```yaml
pull_request:
  branches: [main]
```

Retargeting this PR's base to `feat/delegate-unsubscribe-introspection` (so the diff shows only its own change) means **none of the Rust jobs fire**. The check list shows `license/cla` and nothing else. Flagging it explicitly because a PR with one green check and no red ones reads as passing, and this one has simply not been tested by CI at all.

**What has actually been verified**, locally on `2bca2c0`, which is not a substitute for CI but is not nothing:

- `cargo test --features contract,net,testing,trace` — 127 passed, 0 failed
- `cargo clippy --all-targets --features contract,net,testing,trace -- -D warnings` — clean
- `cargo fmt --check` — clean

### How this resolves

When #98 merges, GitHub retargets this PR to `main` automatically, CI fires, and the normal gate applies. So the sequence is: **#98 merges → this PR's CI runs → then and only then is this mergeable.** It must not be merged on the strength of a local run.

### The alternative, if that ordering is inconvenient

Point this back at `main` to get CI now, at the cost of the diff including all of #98's changes and being much harder to review. I have kept the reviewable diff and accepted no CI, on the assumption #98 lands first — but that is a judgement call, and if the release sequencing changes it should be revisited.

Worth noting separately: **any future stacked PR in this repo hits the same wall.** If stacking becomes a normal pattern here, `ci.yml`'s trigger wants widening. That is a change to a shared CI surface, so I have not made it as a side effect of this work.

[AI-assisted - Claude]

sanity:
## Blocking, found before review: a delegate cannot supply `at`, because it has no clock

Asked to pin down the clock model behind `ScheduleWakeup { at: SystemTime, .. }`. Answering from the code rather than the design note, and the answer is worse than an ambiguity.

### 1. `at` is absolute wall-clock time measured from the Unix epoch

bincode encodes `SystemTime` as **u64 seconds LE + u32 nanos**, relative to `UNIX_EPOCH` — 12 bytes:

```
UNIX_EPOCH + 1000s + 7ns  ->  e8 03 00 00 00 00 00 00  07 00 00 00
```

So it is not a monotonic instant and carries no timezone or origin of its own. It is "seconds since 1970 on somebody's wall clock".

Also worth pinning: **a `SystemTime` before the epoch fails to serialize**, with `SystemTime must be later than UNIX_EPOCH`. That is an encode-time `Err` on the delegate's own outbound message, not a host-side rejection.

### 2. The delegate has no way to obtain the current time

This is the blocking part. `ScheduleWakeup` requires the delegate to compute an absolute instant, and **no clock is reachable from inside a delegate**.

- Delegates compile to `wasm32-unknown-unknown`. `SystemTime::now()` compiles there and **panics at runtime**. Built a probe module and confirmed the panic string is in the binary:

  ```
  $ strings wasmclock.wasm | grep -i "time not implemented"
  time not implemented on this platform
  ```

  The module imports **nothing** — there is no host function it could be calling instead.

- freenet-core registers **17** host functions across the four delegate namespaces (`freenet_delegate_ctx`, `_secrets`, `_contracts`, `_management`). None is temporal. No WASI, no `WasiCtx`, no `add_to_linker`, no `clock_time_get` — greps return nothing in `wasmtime_engine.rs`.

So a delegate cannot express "one week from now". It cannot express *any* `at`, except by hardcoding a constant or echoing a value handed to it from outside.

**That breaks the driving use case in this PR's own description.** River's weekly secret rotation (freenet/river#228) is precisely "wake me in seven days", and it is the thing this primitive exists for.

`WakeupFired { tag }` carries no timestamp either, so a woken delegate cannot learn the current time even at the moment it is woken — it cannot compute the *next* deadline to re-arm with. The recurring case is doubly blocked.

### 3. The persistence claim is not verifiable and has precedent against it

The doc states: *"The host persists the schedule so it survives node restart."* No host implementation exists, so this is a **requirement on freenet-core#3972 asserted as a property**. The precedent is discouraging: `DELEGATE_SUBSCRIPTIONS` is a `LazyLock<DashMap>` with no persistence at all (`native_api.rs:41-43`), so the one comparable piece of per-delegate host state is lost on restart. #98's review found a near-identical claim about subscription replay and had to correct it.

Not asking to solve persistence here — only that the doc stop stating it as fact.

### 4. Clock-jump semantics are unstated

With absolute time and a wall clock: NTP stepping backwards silently delays a wakeup; stepping forwards fires it early or immediately, possibly firing a large backlog at once. Nothing says which, and a relative delay would not have the question.

## Recommendation

**Change `at: SystemTime` to a relative delay, `after: Duration`.** The delegate says "in seven days"; the host adds it to its own clock. Then:

- No clock is needed on the delegate side, so the primitive becomes usable — this is the whole of the blocker.
- No pre-epoch encode failure.
- Clock-jump semantics collapse to the host's own scheduler, which is where the answer belongs.
- `Duration` bincodes to the same 12 bytes (u64 secs + u32 nanos), so the wire size is unchanged.

The cost is that absolute scheduling ("03:00 next Monday") becomes inexpressible. That is not a regression, because it is **already** inexpressible without a clock. If it is wanted later it needs a time host function, which is additive and can land separately — and per `WIRE-FORMAT.md` a host function is the cheaper thing to add later, because it takes no wire tag.

If absolute is preferred instead, then **a clock host function must ship in the same release**, or this variant is dead on arrival.

Recurring drift is the one thing relative loses: a delegate re-arming inside its handler accumulates the handler's latency each cycle. For weekly rotation that is irrelevant; if it ever matters, the fix is a timestamp on `WakeupFired`, which is worth considering now while the shape is still free.

## I have not changed the wire

You said the review is commissioned on `8540b752` and there is nothing for me to do on these three. Changing the field under the reviewer would be exactly the per-code-content staleness we have been avoiding all day. Flagging it here so the review starts with it rather than rediscovering it, and so the decision is yours.

Verification: bincode encoding measured against bincode 1.3.3; the wasm32 panic confirmed by building a probe module and inspecting the binary; the absence of any host clock confirmed by grepping freenet-core's linker registration at `3858a84a4`.

[AI-assisted - Claude]

sanity:
## Correction: my "a delegate has no clock" finding was wrong

Retracting the blocking finding above. **Delegates can read a clock**, and the error was mine.

### What is actually true

- `freenet_stdlib::time::now() -> DateTime<Utc>` exists at `rust/src/time.rs:7`, importing `freenet_time::__frnt__time__utc_now` (`time.rs:18-21`).
- freenet-core registers it in `register_host_functions` (`wasmtime_engine.rs:1892`), and that function is called at **both** instantiation sites (`:836`, `:1443`) — so the `freenet_time` namespace resolves for delegates, not only contracts.
- The #5465 host-clock deprecation does not contradict this; it is contract-only by design.

### How I got it wrong, since the shape matters more than the fact

My probe was sound: `SystemTime::now()` really does compile and panic on `wasm32-unknown-unknown`, and that module really does import nothing. Both facts hold.

**The inference did not.** I enumerated the four `freenet_delegate_*` namespaces, found nothing temporal, and concluded no clock was reachable — but the clock lives in a **fifth namespace**, `freenet_time`, which I never grepped for because it is not delegate-prefixed. I tested the `std` path and generalised to the capability, when the capability arrives by host import. `freenet_stdlib::time` was in the same crate I had open.

Reported as "verified from the code", which made it more load-bearing than it had earned. It is the same failure this workstream keeps finding — a true measurement answering a narrower question than the one asked — arriving this time in my own evidence rather than in something I was reviewing.

### Consequence

The argument that killed `at` is gone. `after: Duration` on the current head was decided on that false premise and is being revised to **`at: u64` — milliseconds since the UNIX epoch** — which is a better answer than either draft:

- Absolute is genuinely expressible now, and a deadline is the natural model for a scheduler.
- `u64` fixes a real defect `SystemTime` has. `SystemTime` serialization **fails** for any pre-epoch value, and `delegate_interface.rs:1442` `.unwrap()`s that result — inside WASM an `unwrap` on `Err` is a trap, so the delegate aborts opaquely rather than returning a `DelegateError`, **taking any sibling message in the same batch with it**. Reachable by ordinary arithmetic (`stored_last_run + interval` with a zeroed stored value). `u64` millis cannot fail to encode.
- It removes the `DateTime<Utc>` / `SystemTime` seam, since `time::now()` returns the former while the field wanted the latter.
- Epoch and precision are in the type instead of in prose.

**Holding the wire change** until the team lead sends the complete finding list, to revise once rather than twice. This head (`70f69ac`, `after: Duration`) is therefore not the shape to review; I will post the new SHA when it lands.

### Already fixed, unrelated to the above

The **L1 body errors** — "tag 8" for variants that are at 9, a "Bumps 0.8.2 → 0.8.3" line when Cargo.toml is 0.10.0 and untouched in the diff, and the "unknown inbound variants are forwarded to the delegate WASM unchanged" claim that reads as a wire tolerance which does not exist — were all in the **pre-rebase body** and were removed when I rewrote it. Verified against the current body: no "tag 8" reference to the new variants, no version claim at all, and the forwarding claim is gone from the body, the doc comments and the CHANGELOG. Worth stating precisely, since squash-merge takes the body and a wire change whose permanent record names the wrong tag is not recoverable.

[AI-assisted - Claude]

sanity:
## All four addressed — current head is `04b5348`, not `70f69ac`

`at: u64` was **never made**; I held it pending the full list, and the list cancelled it. **`after: Duration` stands.** It survives as the host function's parameter type.

### M1 — outbound half converted to a host function

```rust
DelegateCtx::schedule_wakeup(after: Duration, tag: &[u8]) -> Result<(), i64>   // freenet_delegate_management
InboundDelegateMsg::WakeupFired { tag: Vec<u8> }                               // wire, tag 9
```

`OutboundDelegateMsg::ScheduleWakeup` is gone from the wire — `OUTBOUND_VARIANT_COUNT` back to **9**, inbound stays **10**, and the FlatBuffers encoder arm went with it.

I verified the batch argument at source rather than taking it on report. `delegate_interface.rs` serializes the delegate's outbound messages as a single `bincode::serialize(&value).unwrap()` over `Result<Vec<OutboundDelegateMsg>, DelegateError>` — the whole batch in one call, decoded whole by the host. So an unknown outbound variant fails the entire batch, and the sibling `ApplicationMessage` dies with it. Confirmed.

Namespace is `freenet_delegate_management` rather than `_contracts`: this is delegate lifecycle, where `create_delegate` already lives, not a contract-state operation.

### M4 — `tag` bounded

`MAX_WAKEUP_TAG_BYTES = 128`, enforced **before** the host call. A count cap on pending wakeups reads as a bound and is not one while the per-item size is caller-chosen.

### M5 — minimum delay

`MIN_WAKEUP_DELAY = 1s`, documented as a host clamp (clamped **up**). Moving from a deadline to a delay removed the *"deadline in the past"* spelling of the tight-wake-loop hazard, not the hazard: `Duration::ZERO` re-armed inside the handler is the same loop.

### M6 — the context cache during a wakeup

Answered from core, and it turned out to be a **second, independent argument** for `WakeupFired` carrying no context.

`DelegateContextCache` is `DashMap<DelegateKey, DelegateContextEntry>` (`native_api.rs:76`) — keyed **per delegate, not per conversation** — and prunes after `DELEGATE_CONTEXT_TTL`, 10 minutes. So:

- any wakeup worth scheduling far outlives 10 minutes, and the context present at scheduling time is **gone** when it fires;
- if a live context *does* exist inside that window, it belongs to a **different in-flight exchange**, so reading it during a wakeup would be reading another conversation's working state.

There is no coherent value to put in the field. Core's own cache doc says state needing to survive longer belongs in the delegate's secrets, which is where River's rotation already keeps the room list.

### Guards made to fail

| Mutation | Result |
|---|---|
| Tag cap check removed | `an_oversized_tag_is_refused_without_calling_the_host` **fails** ✓ |
| Off-by-one, `>` → `>=` | `a_tag_exactly_at_the_cap_is_not_refused_for_being_too_big` **fails** ✓ |

The off-WASM stub returns `Err(ERR_NOT_IN_PROCESS)`, never `Ok(())`, and that is asserted — a stub reporting success would let a host-side test read "wakeup scheduled" out of a call that scheduled nothing.

### L1

Verified against the **rendered body**, not from memory: no "tag 8" for the new variants, no version claim at all, and the "forwarded to the delegate WASM unchanged" line is absent from the body, the doc comments and the CHANGELOG. The body has since been rewritten again for the host-function shape, so it now describes what actually merges.

130 tests pass, both clippy invocations clean, wasm32 build clean, fmt clean, CI green on `04b5348`. Version unchanged at 0.10.0.

[AI-assisted - Claude]

sanity:

**This bypasses a misconfigured branch protection, not a review or a CI gate.** `freenet-stdlib` main uses classic branch protection with `required_pull_request_reviews` present and `required_approving_review_count: 0` — it blocks every merge while requiring nothing that can satisfy it, and the rules API reports all enforced rules passing, which makes it confusing to diagnose. Recent stdlib merges have gone through the `Freenet Developers` bypass for the same reason. The config is worth fixing.

What the bypass is **not** skipping:

- **All 8 required checks green** at `2bd7b8e867de7838c9d7f39e6435ab48d9a19e60`, the exact commit merged.
- **Five review passes**, final verdict clean, plus a separate targeted check of the post-#128 rebase.

Findings addressed along the way, for the record:

- The outbound half became a **host function** rather than an `OutboundDelegateMsg` variant. Core decodes the delegate's whole outbound batch in one `bincode` call, so an unknown variant failed the batch and took the sibling `ApplicationMessage` down with it — the user's action silently doing nothing, on the rollout direction that happens every time under stdlib-first. A missing host import fails loudly at instantiation instead. That property is *not* currently tested anywhere in either repo; it is recorded as owed on freenet-core#3972.
- `MIN_WAKEUP_DELAY` is now **enforced in the wrapper**, not merely documented. `prepare_wakeup` carries the prologue and `schedule_wakeup` binds its result, so the wiring cannot be deleted without breaking compilation — verified: the same deletion previously shipped green at 133/133. Millisecond saturation is now genuinely exercised rather than merely reachable.
- Several claims about host behaviour were restated as **obligations on freenet-core#3972** rather than present-tense properties: the pending-wakeup count cap, the delay floor, and persistence across restart. None of them exists yet.
- The tag-cap rationale now says plainly that a guest-side check on a guest-controlled import is fail-fast convenience and **can never be a bound**, because the guest controls its own imports — and that this holds for every bound this crate documents, not just this one.

One known cosmetic residue from the rebase: `delegate_host.rs:224` lacks a blank line between two items at the conflict seam. `cargo fmt --check` passes; it is the only trace the resolution left.

[AI-assisted - Claude]

PR #83, opened 2026-07-09
fix: stop leaking every inbound WebSocket message in the WASM WebApi

### LINKED ISSUE #4746: stdlib WASM WebApi leaks every inbound WebSocket message (FileReader + full payload retained forever) — multi-GB River tabs
## Summary

`WebApi::start` in freenet-stdlib's WASM client (`rust/src/client_api/browser.rs`) permanently leaks **every inbound WebSocket message**: the per-message `FileReader` and its `onloadend` closure are `forget()`-leaked, and `FileReader.result` pins the full decoded payload as an `ArrayBuffer` for the life of the tab. Long-lived apps built on `WebApi` (River is the main one) grow without bound — users report River tabs reaching multiple GB in Chrome (system lockup when swap is exhausted) and ~3 GB in Firefox. Because ArrayBuffer backing stores live outside the V8 JS heap, the growth is invisible in `usedJSHeapSize` and shows up only as renderer RSS / tab memory.

## Mechanism

In the `onmessage` handler ([browser.rs L38-L128 @ 75e3a46](https://github.com/freenet/freenet-stdlib/blob/75e3a46/rust/src/client_api/browser.rs#L38-L128)), for every message:

1. A `web_sys::FileReader` is created to read the incoming Blob (the socket uses `binaryType = "blob"`, L164).
2. An `onloadend` closure is created, capturing a clone of the FileReader (`fr_clone`).
3. `onloadend_callback.forget()` (L125) leaks the closure **permanently** — wasm-bindgen never releases it.
4. The leaked closure pins the FileReader; after the read completes, `FileReader.result` holds the entire message payload as an `ArrayBuffer`. Retained forever.

Net effect per inbound message: 1 FileReader + 1 event listener + 1 wasm-bindgen closure + **the full payload bytes**, leaked.

## Measurements (live River tab, headless Chrome 149 via CDP, node on framework, 3 rooms)

- After ~80 min of mostly-idle operation: **506 FileReaders, all `readyState == DONE`, retaining 65.2 MB of ArrayBuffers** (`Runtime.queryObjects(FileReader.prototype)` census).
- FileReader count matches the `JSEventListeners` metric growth 1:1 (~102 → ~500 over the session; DOM nodes flat, detached nodes flat, no WebSocket/BroadcastChannel/MessagePort accumulation — everything else was ruled out).
- Every forced reconnect leaked ~12 more; **every tab-switch-away-and-back leaked ~15 more** (River's `PageBecameVisible` → `RefreshAllRooms` re-GETs the full state of every room, so each tab switch permanently leaks a full copy of every room's state — this is the big accelerant for real users).
- WASM linear memory stayed flat (57.1 MB) and JS heap flat (~2.7 MB) throughout — the leak is entirely in JS-side ArrayBuffers, consistent with users seeing huge "tab memory" while nothing obvious shows in the JS heap.

Growth rate scales with inbound traffic: an active room set plus frequent tab switching reaches GBs in hours-to-days. Browser-agnostic (same leak in Firefox, matching the ~3 GB reports there).

## Fix

Set `binaryType = "arraybuffer"` on the socket and decode `e.data()` synchronously in `onmessage` — this deletes the per-message FileReader/closure entirely (there is no need for the Blob → FileReader async hop). PR incoming.

Two smaller related items, not fixed here:
- The per-connection `forget()`s (`onmessage`/`onerror`/`onopen`/`onclose`, L128-L162) leak ~4 closures per reconnect. Bounded and tiny by comparison, but a `WebApi` that properly owns its closures (dropping them on `Drop`) would clean this up too.
- River-side: `RefreshAllRooms` on every `visibilitychange` (ui `freenet_synchronizer.rs:245-268`) re-fetches full state of all rooms on every tab switch-back. With the leak fixed it's just bandwidth/CPU churn, but it's worth revisiting whether a cheaper liveness check suffices (it also competes with first paint right at tab activation, which users perceive as the tab flashing white before it renders).

## Affected

Any WASM app using `WebApi::start`: River, Atlas UI, node dashboard webapps, etc.

[AI-assisted - Claude]

sanity:
Fixes freenet/freenet-core#4746 (issues are disabled on this repo).

## Problem

`WebApi::start`'s `onmessage` handler read every incoming Blob with a per-message `FileReader` whose `onloadend` closure was `forget()`-leaked. The leaked closure pins the FileReader, and `FileReader.result` pins the full decoded payload, so **every inbound WebSocket message's bytes were retained for the life of the tab**. Long-lived consumers (River) grow to multi-GB tab memory in both Chrome and Firefox.

Measured live against a real node (headless Chrome 149 driven over CDP, 3-room signed-in River tab): 506 leaked FileReaders retaining 65 MB of ArrayBuffers after ~80 mostly-idle minutes; +15 per tab-switch-back (River refreshes all rooms on `visibilitychange`), +12 per reconnect; FileReader count tracks the `JSEventListeners` metric 1:1. Details and full methodology in freenet/freenet-core#4746.

## Fix

Set `binaryType = "arraybuffer"` on the socket and decode `e.data()` synchronously in `onmessage`. This removes the Blob → FileReader async hop entirely, so there is no per-message closure to leak and no FileReader to pin payloads. A non-binary frame (which the old code would have crashed on via unchecked casts) now reports a connection error through the normal error handler.

Behavioral notes:
- Same wire format, same `HostResult` dispatch, same streaming reassembly path — only the frame-decode transport changed.
- Decode now happens in the same microtask as message delivery instead of a FileReader callback; ordering between messages is unchanged (FileReader completions were already FIFO in practice, but ArrayBuffer delivery removes any reliance on that).
- The per-connection `forget()`s (onmessage/onerror/onopen/onclose, ~4 closures per reconnect) remain; they're bounded and small, noted in #4746 as a follow-up.

## Verification

- `cargo check`/`clippy --target wasm32-unknown-unknown --features net` clean; native tests pass; fmt clean.
- The retention mechanism was pinpointed empirically (queryObjects census: FileReaders were the only growing object class, all `readyState DONE`, holding exactly the leaked bytes); this change removes that object class from the path by construction.

[AI-assisted - Claude]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,33 @@
 # Changelog
 
+## [0.8.3] - 2026-07-10
+
+### Fixed
+- **WASM `WebApi` leaked every inbound WebSocket message for the life of
+  the tab.** `WebApi::start`'s `onmessage` handler decoded each incoming
+  Blob with a per-message `FileReader` whose `onloadend` closure was
+  `forget()`-leaked; the leaked closure pinned the `FileReader`, and
+  `FileReader.result` pinned the full decoded payload, so every inbound
+  message's bytes were retained forever. Long-lived consumers (River)
+  grew to multi-GB tab memory in both Chrome and Firefox. The socket now
+  sets `binaryType = "arraybuffer"` and decodes `e.data()` synchronously
+  in `onmessage`, removing the Blob → `FileReader` async hop entirely, so
+  there is no per-message closure to leak and no `FileReader` to pin
+  payloads. Same wire format, same `HostResult` dispatch, same streaming
+  reassembly path — only the frame-decode transport changed. A non-binary
+  frame now reports a connection error through the normal error handler
+  instead of crashing on an unchecked cast. See
+  freenet/freenet-core#4746.
+- **`BorrowMutError` → WASM abort on any malformed or duplicate stream
+  chunk.** The `receive_chunk` `borrow_mut()` in the reassembly `match`
+  scrutinee lived until the end of the `match`, so the error arm's
+  `remove_stream` re-borrow panicked (`BorrowMutError`, aborting the WASM
+  instance) on any malformed or duplicate stream chunk. The borrow is now
+  hoisted out of the scrutinee. `binaryType` is also set before the
+  handlers are installed so the ordering is self-evidently safe rather
+  than relying on `start()` being synchronous. Pre-existing on `main`,
+  surfaced by PR review.
+
 ## [0.8.0]
 
 ### Fixed
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "freenet-stdlib"
-version = "0.8.2"
+version = "0.8.3"
 edition = "2021"
 rust-version = "1.80"
 publish = true
--- a/rust/src/client_api/browser.rs
+++ b/rust/src/client_api/browser.rs
@@ -36,93 +36,81 @@ impl WebApi {
         let reassembly = Rc::new(RefCell::new(super::streaming::ReassemblyBuffer::new()));
 
         let onmessage_callback = Closure::<dyn FnMut(_)>::new(move |e: MessageEvent| {
-            // Extract the Blob from the MessageEvent
+            // Binary frames arrive as ArrayBuffer (`binaryType` is set below and
+            // must stay in sync with this decode path). Do NOT route through
+            // Blob + FileReader here: a per-message FileReader whose `onloadend`
+            // closure is `forget()`-leaked pins `FileReader.result`, retaining
+            // every inbound payload for the life of the tab
+            // (https://github.com/freenet/freenet-core/issues/4746).
             let value: JsValue = e.data();
-            let blob: web_sys::Blob = value.into();
+            let array_buffer = match value.dyn_into::<js_sys::ArrayBuffer>() {
+                Ok(ab) => ab,
+                Err(other) => {
+                    eh.borrow_mut()(Error::ConnectionError(serde_json::json!({
+                        "error": format!("unexpected non-binary websocket message: {other:?}"),
+                        "source": "host response decoding"
+                    })));
+                    return;
+                }
+            };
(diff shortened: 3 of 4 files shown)


sanity:
## Review record (multi-model)

**External (non-Claude):** Codex CLI `review --base main`, run 3× — on the initial diff, after stripping the accidental generated-file churn, and on the final HEAD (f77f788). All passes: no correctness issues found.

**Claude subagents, 3 independent adversarial lenses** (skeptical bug-hunt / browser-API + WebSocket semantics / code-first + downstream-compat), each blind to the others:

1. **[HIGH, found by all three] Latent `RefCell` double-borrow panic** in the `StreamChunk` `Err` arm: the scrutinee's `reassembly.borrow_mut()` temporary lives to the end of the `match`, so the arm's `remove_stream` re-borrow panics (BorrowMutError → WASM abort) on any malformed/duplicate/over-limit chunk from the node. Pre-existing on `main` (identical shape), not introduced here — **fixed in this PR** (fda66d5) by binding the outcome before the match, since it's in the exact block being rewritten.
2. **[Diff hygiene, found by two] Accidental flatbuffers regen** (~4k lines, same failure as #72/#73): `build.rs` regenerates tracked files with the local flatc on every build. **Stripped from this PR** (force-push d04b5c5); follow-up filed as freenet/freenet-core#4747.
3. **[Nit] `binaryType` set after handlers** — safe (start() is synchronous) but moved before handler installation for robustness (fda66d5).

Verified by reviewers, no action needed: binaryType/onmessage ordering safety; text-frame handling (old code panicked via unchecked casts, new path reports a connection error); no new main-thread blocking (decode was always synchronous, now one copy fewer); strict FIFO delivery (the FileReader hop had no cross-message ordering guarantee); no re-entrancy/borrow hazards from user callbacks; API-compatible with all consumers — and the change brings the Rust browser impl into parity with the TS SDK (`binaryType = "arraybuffer"`, websocket-interface.ts:702) and the native impl (synchronous binary decode, regular.rs).

CI: green on the final HEAD except as noted in checks; the earlier wasm-clippy failure was a pre-existing `clippy::question_mark` in `encoding.rs` newly enforced by current stable clippy (main would fail it too) — fixed here (f77f788).

[AI-assisted - Claude]

PR #85, opened 2026-07-20
fix: pass ContractKey code hash through FBS decode instead of re-hashing

sanity:
## Problem

`ContractKey::try_decode_fbs` called `CodeHash::from_code(code_hash.bytes())` on the wire `code` field. That field already carries the computed 32-byte code hash, and `from_code` hashes its input — so decoding produced BLAKE3(BLAKE3(wasm)), a key that never matches the store.

Every FBS encode site writes `key.code_hash().0` (the raw 32 bytes), so the wire contract genuinely is "`code` carries the hash" and passing it through is correct. `DelegateKey::try_decode_fbs` already does it that way.

Blast radius: FlatBuffers clients, which is the default for the TypeScript SDK — browser apps. GET and SUBSCRIBE were unaffected because their `ContractRequest` variants carry only an instance id and never decode `code`; UPDATE is the one variant carrying a full `ContractKey`. Affected apps looked healthy until the first write, which surfaced as a 30s timeout rather than an error.

## What this does

1. **Pass the code hash through** instead of re-hashing it.
2. **Reject a `code` field that is not exactly 32 bytes**, with an error a client developer can act on.

## Why reject rather than tolerate

The schema marks `code` optional, and the TypeScript SDK's `ContractKey.fromInstanceId(...)` emits a **present but zero-length** vector — `pack()` always adds the vector, so this is a shape real first-party producers emit, not a hand-built artifact. The old test fixture in this repo carried exactly that shape.

It is nonetheless safe to reject, because such an UPDATE **has never worked**:

- `try_decode_fbs` is reached from exactly one production site, the UPDATE decode (`client_events.rs`).
- freenet-core gates on `code_blob_stored(key.code_hash())` (`contract/executor/runtime/executor_impl.rs:250`), a pure code-hash lookup.
- UPDATE supplies no contract code — `maybe_defer_upsert(..., code: None, ...)` (`contract.rs:2334`).
- So an empty `code` hashed to BLAKE3 of empty input and failed at the store gate with `"Contract not in store and no code provided"`.

**Wire-compat consequence, stated plainly:** this hard-errors a message shape the TypeScript SDK can produce and the schema permits. But every such message was already failing — later, and with an error pointing at the node rather than the caller. No working client regresses. Clients that work today, including the freechan UI that validated the previous commit end to end, all send a real 32-byte hash. What changes for an affected client is a 30s timeout becoming an immediate, explicit error naming the cause.

The error text now carries the field name, the expected length, the observed length, the reason the hash is required, and a pointer to the real fix. The previous text was `io::ErrorKind::InvalidData`, which stringifies to `"invalid data"`.

**The schema is deliberately unchanged.** Marking `code` `(required)` would be a wire break for the absent case and would not help the empty case anyway, since a required field can still be zero-length.

## The real fix, tracked separately

The honest end state is `Option<CodeHash>` threaded through `ContractRequest::Update`, with the node resolving a `None` from the instance id — which it already knows how to do: `code_hash_from_id` exists and is documented for exactly this ("Used when clients request contracts without knowing the code hash"), and GET and SUBSCRIBE already use it via `lookup_key`. UPDATE is the odd one out only because its wire type carries a full key.

That needs a coordinated stdlib + core change, so it is **freenet/freenet-core#4978**, not this PR. Corroboration that it is the right target: `fdev execute update` sends 32 zero bytes with the comment "the node will look up the actual key" (`crates/fdev/src/commands.rs:419-421`), which is false for UPDATE though true for GET.

## Testing

- `contract_key_code_hash_passes_through_fbs_decode` — the double-hash regression.
- `contract_key_decode_rejects_empty_code_with_actionable_error` — pins **both** the rejection and the message content, since reverting to a bare `try_from` error would leave the rejection green while destroying the reason for the change.
- `contract_key_decode_rejects_absent_code_with_actionable_error` — the absent case keeps its pre-existing behavior; only its message changed, so what is pinned is that the two paths stay consistent.
- `contract_key_decode_rejects_wrong_length_code` — guards the partial fix of special-casing only the empty vector.
- `test_build_contract_update_op_from_fbs` — **rebuilt programmatically.** Its hardcoded byte blob is why the empty-`code` case went unnoticed: reading the test told you nothing and finding it took instrumenting the decoder. It now carries a real 32-byte hash and asserts the hash survives decode, so it covers the double-hash regression too; previously it asserted only the instance id.

Mutation-verified both directions: reverting the error message fails the two message-asserting tests, and reintroducing the double-hash fails both the fixture and the pass-through test (the old blob fixture would not have caught it).

`cargo fmt`, `cargo clippy --all-targets --features net -- -D warnings`, and the full suite (61 tests) are clean.

[AI-assisted - Claude]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,51 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+- **`ContractKey::try_decode_fbs` no longer double-hashes the code hash.** The
+  wire `code` field carries the already-computed 32-byte code hash, but the
+  decoder passed it to `CodeHash::from_code` — the *hashing* constructor —
+  producing `BLAKE3(BLAKE3(wasm))`, a key that never matches the store. Every
+  FlatBuffers `UpdateRequest` failed as a result, which the client surfaced as
+  a 30-second timeout rather than an error. GET and SUBSCRIBE were unaffected
+  because their request variants carry only an instance id and never decode
+  `code`. The regression dates to `844880e` (Nov 2023), which changed a
+  pass-through `CodeHash::new` into `from_code`. Blast radius was FlatBuffers
+  clients, the TypeScript SDK's default, so browser apps: the audience least
+  able to diagnose it.
+
+- **A wrong-length `instance` no longer panics the connection task.**
+  `instance`/`data` are `(required)` in the schema, but the flatbuffers
+  verifier checks that a required vector is present, not that it is the right
+  length. An 8-byte instance passed verification and then hit a
+  `try_into().unwrap()`, panicking with `TryFromSliceError`; nothing catches
+  unwind on that path, so a malformed message from any peer killed that
+  client's connection task. Reachable from UPDATE, GET and SUBSCRIBE; all
+  three now reject with an explicit error.
+
+### Compatibility
+- **A `ContractKey` whose `code` field is absent or not exactly 32 bytes is now
+  rejected at decode.** This is a hard rejection of a shape `common.fbs`
+  permits (`code` is not `(required)`) and that the TypeScript SDK actually
+  emits: `ContractKey.fromInstanceId(...)` produces a present-but-zero-length
+  vector, and the SDK's own test suite builds an `UpdateRequest` that way.
+
+  **No working client regresses.** Such an UPDATE has never succeeded. The node
+  gates an UPDATE on already holding the contract's code blob and probes for it
+  by code hash; an UPDATE supplies no contract code, so a zero-length `code`
+  hashed to `BLAKE3("")` and failed at that gate. What changes is *where* and
+  *how* it fails: previously a 30-second timeout or an opaque
+  `"missing contract: <key>"` pointing at the node, now an immediate
+  `ContractKey.code must be the 32-byte contract code hash; got 0 bytes...`
+  naming the field and the remedy.
+
+  The real fix is for the node to resolve the code hash from the instance id,
+  as GET and SUBSCRIBE already do via `code_hash_from_id` — tracked in
+  freenet-core#4978. Until then, build keys with both parts:
+  `new ContractKey(instance, code)` rather than
+  `ContractKey.fromInstanceId(...)`.
+
 ## [0.8.4] - 2026-07-21
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@
   unchecked-`unwrap` class survives elsewhere on the same UPDATE request -
   `UpdateData`'s `Related*` variants apply base58 decoding to bytes already in
   final form and then unwrap: so an FBS UPDATE carrying one of those variants
-  still panics. Tracked in freenet-core#4983.
+  still panics. Tracked in freenet-core#4996.
 
 ### Compatibility
 - **A `ContractKey` whose `code` field is absent or not exactly 32 bytes is now
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixed
 - **`ContractKey::try_decode_fbs` no longer double-hashes the code hash.** The
   wire `code` field carries the already-computed 32-byte code hash, but the
-  decoder passed it to `CodeHash::from_code` — the *hashing* constructor —
+  decoder passed it to `CodeHash::from_code`: the *hashing* constructor -
   producing `BLAKE3(BLAKE3(wasm))`, a key that never matches the store. Every
   FlatBuffers `UpdateRequest` failed as a result, which the client surfaced as
   a 30-second timeout rather than an error. GET and SUBSCRIBE were unaffected
@@ -21,8 +21,14 @@
   length. An 8-byte instance passed verification and then hit a
   `try_into().unwrap()`, panicking with `TryFromSliceError`; nothing catches
   unwind on that path, so a malformed message from any peer killed that
-  client's connection task. Reachable from UPDATE, GET and SUBSCRIBE; all
-  three now reject with an explicit error.
+  client's connection task. The `ContractKey.instance` field is now
+  length-checked at all three sites that decode it (UPDATE, GET, SUBSCRIBE).
+
+  Note the narrow scope: this covers the `ContractKey` fields only. The same
+  unchecked-`unwrap` class survives elsewhere on the same UPDATE request -
+  `UpdateData`'s `Related*` variants apply base58 decoding to bytes already in
+  final form and then unwrap: so an FBS UPDATE carrying one of those variants
+  still panics. Tracked in freenet-core#4983.
 
 ### Compatibility
 - **A `ContractKey` whose `code` field is absent or not exactly 32 bytes is now
@@ -41,7 +47,7 @@
   naming the field and the remedy.
 
   The real fix is for the node to resolve the code hash from the instance id,
-  as GET and SUBSCRIBE already do via `code_hash_from_id` — tracked in
+  as GET and SUBSCRIBE already do via `code_hash_from_id`: tracked in
   freenet-core#4978. Until then, build keys with both parts:
   `new ContractKey(instance, code)` rather than
   `ContractKey.fromInstanceId(...)`.
--- a/rust/src/client_api/client_events.rs
+++ b/rust/src/client_api/client_events.rs
@@ -503,8 +503,9 @@ impl<'a> TryFromFbs<&FbsContractRequest<'a>> for ContractRequest<'a> {
                     // Extract just the instance ID - GET only needs the instance ID,
                     // not the full key (which may not be complete on the client side)
                     let fbs_key = get.key();
-                    let key_bytes: [u8; 32] = fbs_key.instance().data().bytes().try_into().unwrap();
-                    let key = ContractInstanceId::new(key_bytes);
+                    let key = crate::contract_interface::key::instance_id_from_fbs(
+                        fbs_key.instance().data().bytes(),
+                    )?;
                     let fetch_contract = get.fetch_contract();
                     let subscribe = get.subscribe();
                     let blocking_subscribe = get.blocking_subscribe();
@@ -541,8 +542,9 @@ impl<'a> TryFromFbs<&FbsContractRequest<'a>> for ContractRequest<'a> {
                     let subscribe = request.contract_request_as_subscribe().unwrap();
                     // Extract just the instance ID for Subscribe
                     let fbs_key = subscribe.key();
-                    let key_bytes: [u8; 32] = fbs_key.instance().data().bytes().try_into().unwrap();
-                    let key = ContractInstanceId::new(key_bytes);
+                    let key = crate::contract_interface::key::instance_id_from_fbs(
+                        fbs_key.instance().data().bytes(),
+                    )?;
                     let summary = subscribe
                         .summary()
                         .map(|summary_data| StateSummary::from(summary_data.bytes()));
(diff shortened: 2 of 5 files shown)


botBehavior:
﻿Independent reproduction and verification of this fix, from someone who hit the bug before
finding the PR.

**Symptom.** A browser app using `@freenetorg/freenet-stdlib` could GET contract state fine but
every `UpdateRequest` failed. The client surfaced it as a 30s timeout rather than an error, so it
read as a network problem.

**Cause, as this PR says.** `try_decode_fbs` calls `CodeHash::from_code` â€” the *hashing*
constructor â€” on a value that is already a code hash, so the decoded `ContractKey` carries
`blake3(code_hash)`.

Confirmed arithmetically rather than by inference: the node logged
`key_code_hash = <instance id>`, and for our contract with empty parameters
`blake3(code_hash) == instance_id` exactly. That is the signature of the double hash, not a
coincidence.

**GET succeeds because it never needs the code**, which is why an affected app looks healthy
until the first write.

**Verified end to end.** With this branch patched in, on a node built from it: FlatBuffers
`UpdateRequest` succeeded and the contract state went 4 â†’ 6 messages. Reverting to stock stdlib
on the same node, same contract, minutes apart:

```
flatbuffers (TS SDK)   UPDATE -> FAILED: Request timeout    7 messages, unchanged
native/bincode         UPDATE -> OK                         7 -> 8, persisted
```

So the blast radius is specifically **FlatBuffers clients**, which is the default for the
TypeScript SDK â€” i.e. browser apps, the audience least able to diagnose it. Rust clients using
`?encodingProtocol=native` are unaffected, which is the workaround we shipped on in the meantime.

**One note for anyone testing this branch.** It declares version `0.8.3`. If you point a
`[patch.crates-io]` at it while building a `freenet-core` that requires `0.8.4`, Cargo
**silently ignores the patch** and you end up testing stock stdlib while believing you tested the
fix â€” we did exactly that and briefly concluded the fix didn't work. Worth checking
`cargo tree -p freenet | grep stdlib` shows a *path* dependency before drawing conclusions. Not a
problem with the PR, which merges cleanly; just a trap for the tester.

**Unrelated but adjacent, in case it saves someone else the confusion:** `fdev execute update`
fails against a healthy contract for a *different* reason â€” it has only a base58 instance id and
sends 32 zero bytes as the code hash (`crates/fdev/src/commands.rs:421`, and the same at `:469`
and `:522`). That is on the native wire, so it fails even with this fix applied. Happy to open a
separate issue if useful.

sanity:
## Review: the diagnosis is right, the implementation over-tightens an optional wire field

Reviewed because a user (ofansifkapital-xmpp) validated `61c0fbe0` end-to-end against a third-party freechan TypeScript UI and asked for merge + publish. The core insight holds up, but this isn't mergeable as-is and I don't think the failing check is incidental.

### Confirmed: the double-hash bug is real, and this is the right direction

Every FBS **encode** site writes `builder.create_vector(&key.code_hash().0)` — the raw 32-byte code hash (`client_events.rs:1014`, `:1111`, `:1128`, and the rest). So the wire contract genuinely is "`code` carries the 32-byte hash", and `try_decode_fbs` calling `CodeHash::from_code` on it produced BLAKE3(BLAKE3(wasm)). Passing through is correct, and matching `DelegateKey::try_decode_fbs` is the right precedent.

### The failing check is caused by this PR, not inherited

I ran `test_build_contract_update_op_from_fbs` at both commits:

- parent `8b537021` — **passes**
- this PR `61c0fbe0` — **fails**, `DeserError { cause: "invalid data" }`

So it's a regression introduced here, not a pre-existing red.

### Root cause: the fixture sends an EMPTY code field, and the schema allows that

I instrumented the decoder to print what the fixture actually carries:

```
PROBE: code field present=true len=Some(0)
```

Zero-length, not 32. And `schemas/flatbuffers/common.fbs` makes that legitimate:

```
table ContractKey {
  instance:ContractInstanceId(required);
  code:[ubyte];            // <-- deliberately NOT required
}
```

`instance` is marked `(required)`; `code` is not. So a client addressing an UPDATE by instance id alone, with no code hash, is doing something the schema explicitly sanctions.

Before: `CodeHash::from_code(&[])` returned BLAKE3 of the empty input — a well-formed but meaningless hash. Wrong, but non-fatal, and the test never asserted on it (it only checks `encoded_contract_id()`, which comes from `instance`).
After: `CodeHash::try_from(&[])` is a length mismatch, so the whole request is rejected.

The net effect is that an **optional** schema field becomes **required-and-exactly-32-bytes**. That is a wire-compatibility break for any client that omits or empties it — and since the fixture in our own test suite does exactly that, at least one such producer was contemplated.

### Why the end-to-end validation doesn't cover this

The reported success is real evidence, but it exercises the path where the client *does* send a genuine 32-byte hash — precisely the case the double-hash bug broke. It cannot tell us anything about the empty-code case, which is the one this change newly rejects. Both can be true at once.

### Suggested shape

Pass through when the field is 32 bytes; keep the empty/absent case working rather than converting it to a hard error. Concretely: treat a 32-byte `code` as the hash, and decide a defined behavior for empty/absent — either preserve today's lenient result or thread an `Option<CodeHash>` through — instead of failing the request. Worth checking what the node actually does with the decoded code hash on the UPDATE path: if it resolves the contract by instance id, the code hash may not need to be authoritative there at all, which would make leniency clearly correct.

Note the pre-existing `.ok_or_else(|| "ContractKey missing code hash")` is also stricter than the schema (it rejects an *absent* `code`), so that inconsistency predates this PR and is worth resolving in the same change.

Also: this is still marked **draft**. Whatever the resolution, this is a wire-boundary change and wants a Full review plus green CI before publish, per the stdlib-first release policy (stdlib must merge AND publish before any core PR consumes it — no `[patch.crates-io]` overrides).

### Merge/publish sequencing, for whoever picks this up

Merging and publishing are safe to decouple from any contract or delegate migration. Both consumers commit lockfiles (River pinned at 0.8.3, freenet-core at 0.8.4), so a newly published version reaches nobody until a pin is deliberately changed. And the consumer that actually needs this fix is **freenet-core** (the node does the decoding); bumping core's stdlib rebuilds the node binary and does **not** rebuild `room_contract` or delegate WASM, so no keys change and no migration entry is required.

[AI-assisted - Claude]

sanity:
---

## Update: review findings addressed (`0ae33fc`)

### The error message stated something false

It said the node "resolves the contract's WASM by code hash". **It does not** — resolution is by **instance id**: `ContractStore::fetch_contract` recovers the hash from `key_to_code_part[key.id()]`, the module cache keys on `ContractKey` whose `Hash`/`Eq` use only the instance, and the state stores key on `as_bytes()`, which is instance bytes. The hash is load-bearing at exactly one place: the gate asking whether this node already holds the code blob.

As written it told developers the hash is fundamentally required — the opposite of what freenet-core#4978 argues, and #4978 is the fix it points them at. An error that confidently misstates how the node works is worse than the opaque `"invalid data"` it replaces.

It also cited `"Contract not in store and no code provided"`, a node-side `tracing::error!` no client ever sees; clients get `MissingContract` → `"missing contract: {key}"`. Both corrected; em-dash removed from the user-visible text.

### A wrong-length `instance` panicked the connection task

`instance`/`data` are `(required)`, but the flatbuffers verifier checks **presence, not length**. An 8-byte instance passed verification and hit `try_into().unwrap()`, panicking with `TryFromSliceError`. Nothing catches unwind on that path and `panic = "abort"` is not set, so a malformed message from any peer killed that client's connection task.

Fixed at all three sites that decode an instance from the wire — UPDATE, GET (`client_events.rs:506`) and SUBSCRIBE (`:544`) — through one shared `instance_id_from_fbs`. Kept as one change deliberately: splitting it would leave two of three known-panicking paths open in a PR whose subject is rejecting malformed keys gracefully.

### The two test suites pinned the same bytes in mirror

The TypeScript suite's `EXPECTED_UPDATE_REQ` is byte-for-byte the blob this PR deletes from the Rust side. After this change `npm test` still asserts as correct an encoding the Rust decoder hard-errors on, and stays green — both suites run in CI, so the disagreement was invisible. That mirror-pinning is exactly how the original double-hash bug survived.

Now pinned from Rust (`typescript_sdk_instance_only_update_is_rejected_with_guidance`), asserting the **actionable message** and not just the rejection, with a matching note on the TypeScript array so the two must be updated together.

### Restored a wire-format pin the programmatic rebuild removed

Rebuilding the update fixture programmatically was right for readability and wrong for coverage: it encodes with the same generated code it decodes with, so a `common.fbs` reorder of `instance` and `code` moves both sides together and stays green while real clients break. After that rebuild, nothing decoded a `common.ContractKey` from fixed bytes at all.

`contract_key_wire_format_is_stable` keeps the programmatic build and adds a checked-in blob asserting a **successful** decode of both fields, following `delegate_interface.rs`'s `*_wire_format_is_stable`. Distinct from the TypeScript pin, which covers the reject path only.

### Regression origin

`844880e` (Nov 2023) is the commit that introduced this: the decoder was a pass-through (`CodeHash::new`) before it changed to `from_code`.

### CHANGELOG

Added, with a `### Compatibility` section — this hard-rejects a shape the schema permits and the TypeScript SDK emits, and the reasoning otherwise lived only here where no consumer would see it.

### Mutation-verified

- Simulating a `common.fbs` field swap (decode `code` as the instance) fails 8 tests including the new wire-format pin.
- Restoring the `try_into().unwrap()` fails 2 tests.
- Reverting `code_hash_error` to `"invalid data"` fails 4, including the TypeScript cross-language pin.

### Follow-up, not done here

Making the TypeScript `UpdateRequest` constructor reject a key whose `code` is not 32 bytes is #4978 territory — a public-API change to the SDK, deliberately out of scope. Skipped 31/33-byte boundary tests: `try_from` is a plain `len() != 32` equality check, so they exercise the same branch as the existing 16-byte case.

`rust/Cargo.toml` deliberately untouched at `0.8.4`.

[AI-assisted - Claude]

sanity:
## Re-review round addressed (`6e7f0ad`)

**1. GET and SUBSCRIBE are now pinned.** You were right that this was the important one — the short-instance test reached the helper only through `ContractKey::try_decode_fbs`, i.e. UPDATE, so reverting either of the other two sites left the whole suite green. Added request-level tests that build a real GET and SUBSCRIBE `ClientRequest` with a wrong-length instance, plus a well-formed case so the guard is shown to reject the wrong size without breaking the right one.

**Verified by reverting each site, not by reasoning:**

| reverted to `try_into().unwrap()` | result |
|---|---|
| `client_events.rs` GET site | 1 test fails |
| `client_events.rs` SUBSCRIBE site | 1 test fails |
| `key.rs` UPDATE site | 2 tests fail |

**2. CHANGELOG claim narrowed.** It said all three paths now reject with an explicit error; that is true only for the `ContractKey` fields. Now states the `ContractKey.instance` scope explicitly and says plainly that the same class survives on the same UPDATE request via `UpdateData`'s `Related*` variants, with a pointer to the follow-up.

**3. TS pin comment softened.** It asserted enforcement that does not exist — the Rust side holds a copy, so editing the TypeScript array fails nothing here, and `include_str!` across to `typescript/` is unavailable because `rust/` is the published package root. Reworded to describe the convention rather than claim a mechanism.

**4. Em-dashes swept** from the lines this PR adds (pre-existing text left alone).

## Follow-up filed: freenet-core#4996

The same bug class on the same request, deliberately not pulled in here:

- `contract_interface/update.rs:324, :331, :341` apply `ContractInstanceId::from_bytes` — a **base58 decoder** — to bytes already in final form, then `.unwrap()`. Structurally the identical "wrong transformation applied to an already-correct value" defect as the double-hash this PR fixes, two lines downstream of it (`client_events.rs:538`). Same shape at `update.rs:139` for PUT related contracts.
- `client_events.rs:766-771` unwraps `cipher`/`nonce` length conversions — `(required)` but length-unchecked, which is this PR's own argument.
- `update.rs:348` is `_ => unreachable!()` on a union whose generated verifier ends in `_ => Ok(())` — the same discriminant panic already fixed for `ContractRequestType`.

All citations verified against the checked-out code before filing.

`rust/Cargo.toml` still untouched at `0.8.4`.

[AI-assisted - Claude]

PR #86, opened 2026-07-22
feat: add DelegateRequest::RegisterDelegateWithPredecessors for secret copy-forward

sanity:
## Problem

There is no wire primitive for the consent-gated, one-click delegate migration flow (freenet-core#4117, freenet-core#2776). When a delegate's WASM changes its key changes, so the new generation starts with an empty secret namespace and users lose their delegate-held data unless secrets can be forwarded from the retired generation(s) into the new one. This PR adds the client-request primitive that carries the migration intent to the node.

## Approach

Add a **new** `DelegateRequest` variant rather than adding a field to the existing `RegisterDelegate`:

```rust
RegisterDelegateWithPredecessors {
    delegate: DelegateContainer,
    cipher: [u8; 32],
    nonce: [u8; 24],
    predecessors: Vec<DelegateKey>,
}
```

### Why a new variant, not a new field (the compat rationale)

`DelegateRequest` crosses the client↔node boundary as **bincode** (`EncodingProtocol::Native`). Bincode encodes an enum variant as a **4-byte little-endian `u32` discriminant** followed by the fields in declaration order. Two consequences drive the design:

- **Adding a field to `RegisterDelegate` is a wire break.** It changes that variant's byte layout, so any already-deployed client speaking the old shape fails to round-trip — the v0.2.11 break class.
- **Appending a whole new variant as the LAST arm is additive.** The tags of `ApplicationMessages` (0), `RegisterDelegate` (1) and `UnregisterDelegate` (2) stay **byte-for-byte identical**; the new variant takes tag **3**. Old clients that never emit it are unaffected; the node accepts it additively. An old node that receives the new variant fails that one request with a clean `DeserializationError` and keeps the connection alive (confirmed by external review).

`predecessors` is a **list** because a user may skip one or more generations between the version they last ran and the current one, so several prior namespaces can need forwarding in a single registration. The node-side copy-forward (implemented separately in freenet-core#4117) is:
- **consent-gated client-side** — the node executes no old delegate WASM; it copies only the already-sealed LOCAL-scope secret bytes it holds for a predecessor,
- **best-effort over unknown predecessors** — the server silently ignores any predecessor key it does not hold, and
- **idempotent per `(predecessor → successor)` pair**.

`cipher`/`nonce` mirror `RegisterDelegate` purely for field-shape parity; the node has ignored both since freenet-core#4140 (secrets are sealed client-side).

### Flatbuffers path intentionally not extended

The `.fbs` schema and generated code are untouched and the `DelegateRequest::try_decode_fbs` known-variant arms are unchanged. The Rust consumers of this feature (river CLI via `regular.rs`, river UI via `browser.rs`) both `bincode::serialize` their requests, so the feature travels on the **Native** path. A flatbuffers (`EncodingProtocol::Flatbuffers`, TS SDK) client cannot encode a union member absent from the schema, so it cannot construct this variant at all. This also keeps the PR clear of the finicky flatc regen handled separately by `fix/4747-stdlib-flatc-regen`.

## Wire-format pin tests (byte-freeze + provenance)

`DelegateRequest` had **no** wire-format pin before this PR — a variant reorder would have shipped undetected. `wire_format_is_frozen` now freezes the **complete** bincode byte vector of all four variants with non-trivial payloads (a real `ApplicationMessage` + params; a multi-element `predecessors` list), asserts the four tags are exactly 0/1/2/3, and additionally **deserializes** each frozen vector to prove an old-format byte stream still decodes into the expected variant.

**Provenance:** the three pre-existing variants' frozen vectors were generated from **origin/main @ 8b53702** (the shipped `0.8.3` format) via a throwaway generator and confirmed byte-identical to this branch, so the pin anchors to what old clients actually speak. The fourth vector (`RegisterDelegateWithPredecessors`) was generated from this branch. This is stated in a comment in the test.

## Reachable-panic hardening (both client-request fbs unions)

Pre-existing, surfaced by review on the exact line this PR's rationale cites. Both `DelegateRequest::try_decode_fbs` and `ContractRequest::try_decode_fbs` matched their generated union type with an `unreachable!()` catch-all — but it **is** reachable: the generated verifier accepts any discriminant it doesn't recognize (`_ => Ok(())`), and the union type field is a raw `u8` the public (TypeScript) builder can set to any value, so a single crafted request reached the catch-all and **panicked the connection handler** (a DoS). Both now return a clean per-request `DeserializationError` naming the unknown discriminant.

Two regression tests build a `ClientRequest` buffer with an out-of-range discriminant (99) via the generated flatbuffers builders and assert a clean `Err` (no panic) — one for each union.

## Testing

- `wire_format_is_frozen` — complete-byte freeze + tag pins + deserialize-old-bytes for all four variants.
- `register_with_predecessors_tag_is_three` — the new variant serializes at tag `3u32` LE.
- `register_with_predecessors_round_trips` — bincode round-trip including a multi-element `predecessors` list.
- `key_returns_the_new_delegate` — `key()` returns the newly-registered delegate's key.
- `fbs_decode_rejects_unknown_discriminant` / `fbs_decode_rejects_unknown_contract_discriminant` — crafted unknown-discriminant buffers return `Err`, not a panic.

`cargo fmt`, `cargo clippy` (x86_64 + wasm32, and `--all-targets`), and the full `cargo test --workspace --features contract,net,testing,trace` suite are green. Bumps `freenet-stdlib` 0.8.3 → 0.8.4.

## Review lineage

External Codex pass: compat verdict clean (existing bytes unchanged; old-node failure is a clean per-request `DeserializationError` with the connection kept alive). It raised two findings, both addressed here: **P1** — freeze complete byte vectors for all four variants (was tag-only for the complex ones) + deserialize the frozen old vectors, anchored to origin/main; **P2** — replace the reachable `unreachable!()` in the fbs decode, and (per maintainer direction) fold in the identical `ContractRequest` sibling in the same file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014wJpXbgCrnkSKgkpmnLGwQ

[AI-assisted - Claude]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## [0.8.4] - 2026-07-21
+
+### Added
+- **`DelegateRequest::RegisterDelegateWithPredecessors`** — a new client
+  request variant for registering a delegate while requesting a one-shot,
+  node-side copy-forward of the LOCAL-scope secrets belonging to one or more
+  retired delegate generations (`predecessors: Vec<DelegateKey>`) into the new
+  delegate's namespace. This is the wire primitive for consent-gated,
+  one-click delegate migration (freenet-core#4117, freenet-core#2776): the
+  node copies already-sealed secret bytes only, performing no execution of any
+  old delegate WASM. The copy-forward is best-effort over unknown
+  predecessors and idempotent per `(predecessor -> successor)` pair. `cipher`
+  and `nonce` mirror `RegisterDelegate` for field-shape parity and are ignored
+  by the node (as they have been since freenet-core#4140).
+
+  The variant is **appended last** (bincode tag `3`, one past
+  `UnregisterDelegate`), so the encodings of `ApplicationMessages`,
+  `RegisterDelegate`, and `UnregisterDelegate` are byte-for-byte unchanged and
+  older clients keep working. New wire-format pin tests lock all four variant
+  tags to guard against a future reorder. The flatbuffers
+  (`EncodingProtocol::Flatbuffers`) client path is intentionally not extended;
+  the Rust consumers of this feature use the bincode `Native` path.
+
 ## [0.8.3] - 2026-07-10
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,10 +18,21 @@
   The variant is **appended last** (bincode tag `3`, one past
   `UnregisterDelegate`), so the encodings of `ApplicationMessages`,
   `RegisterDelegate`, and `UnregisterDelegate` are byte-for-byte unchanged and
-  older clients keep working. New wire-format pin tests lock all four variant
-  tags to guard against a future reorder. The flatbuffers
-  (`EncodingProtocol::Flatbuffers`) client path is intentionally not extended;
-  the Rust consumers of this feature use the bincode `Native` path.
+  older clients keep working. New wire-format pin tests freeze the complete
+  bincode byte vector of all four variants (the three pre-existing ones
+  anchored to the shipped `0.8.3` format) to guard against a future reorder or
+  nested-encoding change. The flatbuffers (`EncodingProtocol::Flatbuffers`)
+  client path is intentionally not extended; the Rust consumers of this feature
+  use the bincode `Native` path.
+
+### Fixed
+- **`DelegateRequest::try_decode_fbs` no longer panics on an unknown union
+  discriminant.** The generated flatbuffers verifier accepts any
+  `DelegateRequestType` discriminant it doesn't recognize (`_ => Ok(())`), and
+  the union type field is a raw `u8` a client can set to any value, so a
+  crafted request reached an `unreachable!()` and took down the connection
+  handler. It now returns a clean per-request `DeserializationError` naming the
+  unknown discriminant.
 
 ## [0.8.3] - 2026-07-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,13 +26,14 @@
   use the bincode `Native` path.
 
 ### Fixed
-- **`DelegateRequest::try_decode_fbs` no longer panics on an unknown union
-  discriminant.** The generated flatbuffers verifier accepts any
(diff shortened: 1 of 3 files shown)


sanity:
## Consolidated review record (Full tier — wire-format surface)

Reviewers: **Codex** external pass plus two independent Claude lenses (**wire-compat**, **consumer-impact** across the whole ecosystem), each blind to the others.

### Verified (the load-bearing properties)
- **Existing bincode bytes unchanged** — proven on independent grounds (bincode-1 positional tags; no enum attr/borrow/shared-type changes; origin/main variant order identical) and now **pinned by complete frozen byte vectors for all four variants**, with the three old vectors generated from origin/main @ 8b53702 and confirmed byte-identical (provenance in-test). Frozen vectors are also deserialized, proving old byte streams decode.
- **Old-node failure mode**: a 0.8.4 client sending tag 3 to an old node gets a clean per-request `DeserializationError` with the connection kept alive — usable as the capability signal. Corollary (consumer lens): **no client may emit the new variant unconditionally**; deploy ordering is stdlib-publish → core-merge → core-release → only then emitting clients.
- **Consumer sweep**: `#[non_exhaustive]` predates this PR, so zero compile breaks anywhere; River/Delta/Atlas/fdev only construct old variants; core's wildcard arms handle the variant gracefully pre-handler (the three match sites needing explicit arms are relayed to the core PR); TS/flatbuffers clients structurally cannot emit it (schema intentionally not extended).

### Findings addressed
- Pin tests upgraded from tag-only to complete-byte freeze (Codex P1) — `a52621a`
- **Pre-existing reachable panic fixed**: the fbs decode catch-alls (`unreachable!()`) for BOTH `DelegateRequest` and `ContractRequest` were reachable by a crafted client writing an unknown union discriminant (the generated verifier accepts them) — now clean deserialization errors, each pinned by a discriminant-99 test — `a52621a`/`a723f1d`

CI green on the final head (`a723f1d`) across both arches + TS SDK. Merging; publishing 0.8.4 via `rust-v0.8.4` (safe pre-handler per the consumer verdict — zero blast radius until core opts in).

Part of freenet-core#4117 (reopened) / freenet-core#2776.

[AI-assisted - Claude]

PR #87, opened 2026-07-27
fix: deliver the buffered response when recv()'s stream channel closes first

sanity:
## Problem

`test_send_oversized_rejected_before_streaming` fails intermittently — **6 runs out of 25 (24%)** of the full suite on a loaded 16-core machine, while passing **40/40** run alone, which is why it reads as an ordinary timing flake.

It is not. The test is correct; it was catching a real bug in `WebApi::recv()`.

```
panicked at rust/src/client_api/regular.rs:757:9:
expected a TotalChunksTooLarge error from the send-path guard,
got: client error: comm channel between client/host closed
```

`recv()` selected over `response_rx` and `stream_rx`, treating either channel returning `None` as a terminal `ChannelClosed`. But `request_handler` sends its final `HostResult` on `response_tx` and *only then* returns, which drops both senders:

```rust
let _ = response_tx.send(Err(error)).await;   // real error queued
// ... task returns here, dropping stream_tx  // stream_rx now closed
```

So whenever the caller is scheduled after the handler has finished — the normal case under load — both arms are ready at the same instant, and `tokio::select!` picks between ready arms **at random**. Roughly half those interleavings took the closed-`stream_rx` arm and threw away the real result sitting in `response_rx`.

**This is user-visible, not a test artifact.** The last message before teardown is the most diagnostic one a client ever gets, and it was being replaced by a generic channel error at random.

**Review found the same defect twelve lines above**, in `Stream::poll_next` — the other public consumption path — where it returned `Poll::Ready(None)` as soon as `response_rx` closed. `poll_recv` returns `Ready(None)` only when closed *and* drained, which is exactly the teardown state, at which point a `WsStreamHandle` may still be queued and assemblies may still be in flight. Same shape, but deterministic rather than 50/50, so it discarded a **complete streamed response** every time.

## Approach

Report a closed connection only once **both** channels are exhausted: if one closes, drain the other before giving up. Applied to `recv()` and to `poll_next`.

Because that fall-through made `poll_next`'s `stream_rx` arm the only way the stream can ever terminate, that arm now also requires that no further response can arrive. It uses the response arm's verdict **from the same poll** rather than a later `is_closed()` check — `is_closed()` means "every sender dropped", not "drained", so between the two the handler could queue its final result and drop both senders, ending the stream with that result still buffered. `Ready(None)` from `poll_recv` means closed and drained, observed atomically.

`recv()`'s `select!` is deliberately **not** `biased`. An earlier revision added it; review showed the justification was wrong. Capacity-1 `response_tx` gives lockstep, not fairness — the handler stays exactly one response ahead, so a prioritised response arm is ready on every call and a queued handle is never picked up, buffering its whole payload in the unbounded chunk channel while never being delivered. With both `None` branches in place, neither ordering can lose a value anyway.

`recv()` is now documented as **not cancellation-safe** (dropping the future mid-assembly loses the response). That was already true; it was simply never stated on a public API. `request_handler` documents that both senders must drop together, since each consumption path treats one channel closing as "check the other".

## Testing

Seven tests, each reproducing an interleaving **directly** rather than racing for it — channels are put in the target state before the call. No sockets, no sleeps, and no dependence on which branch a combinator polls first: the "must still be pending" assertions drive the poll explicitly through a noop waker. Every test whose regression mode is "never resolves" is bounded, so that fails the test instead of hanging CI.

- `recv_delivers_buffered_response_when_stream_channel_closes_first` — the original regression.
- `recv_delivers_queued_stream_after_response_channel_closes` — the mirror, recovering a complete streamed response.
- `recv_keeps_waiting_for_responses_after_the_stream_channel_closes` — a closed stream channel must not abandon a live response channel.
- `poll_next_delivers_queued_stream_after_response_channel_closes` — the same precedence through the `Stream` impl, which previously had no coverage at all.
- `poll_next_ends_only_when_nothing_is_left_anywhere` — both halves of the new terminal condition.
- `poll_next_waits_for_an_assembly_still_in_flight` — the third part of that condition, which the other fixtures cannot reach because their payloads assemble on the first poll.
- `recv_reports_closed_only_when_both_channels_are_exhausted` — `ChannelClosed` is still returned once nothing is left, so the above cannot be satisfied by never reporting closure.

The two tests where both `select!` arms are ready loop 200×, so a revert fails deterministically instead of about half the time.

**Mutation-verified: nine mutants across every fixed branch and guard conjunct, each killed by a targeted test, none surviving.** Four of these pins exist only because mutation testing showed earlier rounds left branches unpinned. One round-3 guard turned out to be *unpinnable by construction* (it defended a cross-thread window no deterministic test can build); rather than ship an unpinned defensive check, `poll_next` was restructured to carry the response arm's verdict, which removes the window instead of narrowing it — and is itself pinned.

| | before | after |
|---|---|---|
| full suite, sequential | 19/25 (6 failures) | **140/140** |
| full suite, 4× parallel contention | — | **168/168** |
| oversized test alone | 40/40 | 40/40 |

`cargo fmt --check` and `cargo clippy --all-targets --features net -- -D warnings` are clean.

## Known residual, not addressed here

At teardown, a queued handle whose chunk sender already died assembles to `StreamError::Truncated`; when a real error is buffered at the same time, the random pick surfaces the truncation instead about half the time. Pre-existing, narrower than the bug fixed here, and the obvious lever (ordering the arms) trades it for the starvation described above.

## Not in scope

Surfaced during review, deliberately left alone rather than smuggled in:

- `recv_stream()` maps a closed channel straight to a generic error while the real one sits unread in `response_rx`. Same symptom, no data loss.
- `handle_response_payload` does `response_tx.send(...).await` on a capacity-1 channel from inside the handler's `select!` loop, so a client that calls `send()` twice without an interleaving `recv()` can wedge both sides. This is the `.send().await`-in-an-event-loop pattern freenet-core's bug-prevention rules call out.

Also worth noting for maintainers: `rust/build.rs` re-runs `flatc` on every build, so a contributor whose local flatc differs from the one that produced the checked-in files silently rewrites `rust/src/generated/*.rs`. That churn rode along on this branch's first commit and has been dropped; the same thing happened in #72 and was reverted in #73.

Found while reviewing #85, which is unrelated and untouched by this branch.

[AI-assisted - Claude]


sanity:
## Review record

Four rounds of independent blind review ran on this branch. Full tier, since the change is in async/`select!`/channel-shutdown territory on a public client API. Lenses: **skeptical** (adversarial bug hunt), **code-first** (read the code before the description), **testing** (coverage and vacuity), and **async failure-mode** (hunt for the same shape elsewhere in the crate). Each round re-ran against the new head, because the code changed materially every time.

What the rounds actually caught, in order:

1. **Round 1** — 2,052 lines of accidental flatbuffers regen riding along (dropped); the same defect unfixed in `Stream::poll_next`, plus a comment asserting the opposite (fixed); the fix's own recovery branch unpinned, with a surviving mutant (pinned); a test whose doc comment described a mirror case it never constructed (rewritten).
2. **Round 2** — the `biased;` added in round 1 was wrong: capacity-1 `response_tx` gives lockstep, not fairness, so prioritising responses can starve the stream arm indefinitely while its chunks pile up in an unbounded channel. No test could pin `biased;` either. Removed. Also: `poll_next`'s terminal condition was one-sided, and my fall-through had made it the *only* way the stream can end.
3. **Round 3** — the `is_closed()` guard added in round 2 rests on a claim tokio does not make (`is_closed()` means "senders dropped", not "drained"), leaving a TOCTOU window where the stream ends with the final response still buffered. The first attempt to patch it with `is_empty()` turned out to be unpinnable by construction, so `poll_next` was restructured to carry the response arm's own verdict — which removes the window instead of narrowing it, and *is* pinnable.
4. **Round 4** — clean. No High or Medium. Four documentation-level Lows, all applied.

**Mutation testing, not assumed:** nine mutants across every fixed branch and every guard conjunct, each killed by a targeted test, none surviving. Four of the seven tests exist specifically because mutation testing showed an earlier round had left a branch green under revert.

Flake verification on the machine that produced the original 24% failure rate: **140/140** sequential full-suite runs and **168/168** under 4× parallel contention, against 19/25 before.

Two pre-existing issues were found and deliberately left out of scope rather than smuggled in — they are listed in the PR description, along with a note about `build.rs` regenerating `rust/src/generated/*.rs` on every build, which is how the regen churn got in here and in #72.

Auto-merge is armed (squash). It needs a code-owner approval to land — branch protection has `require_code_owner_reviews`, which I can't self-provide.

[AI-assisted - Claude]

PR #88, opened 2026-07-27
fix: stop the flatbuffers decode boundary panicking on client input

sanity:
## Problem

Twelve sites on the FlatBuffers decode path either panicked the connection task or silently produced a wrong value, on input a client can actually send — eleven on the request-decode path, plus one encode site found during review. These were found by sweeping every `try_decode_fbs` impl after #85 fixed one instance of the first shape below; the sweep is exhaustive over `.bytes()` against every `try_into` / `try_from` / `copy_from_slice` / `from_bytes` / indexing form.

**Every one of these is FlatBuffers-only.** The native path decodes the same Rust enums via bincode and never reaches `try_decode_fbs`, and first-party tooling (fdev, riverctl, River UI) uses `encodingProtocol=native`. That is both why the class went unnoticed for years and why it lands entirely on third-party and browser clients — the same population that reported the original double-hash.

### Shape 1: a base58 decoder pointed at raw bytes (4 sites)

`RelatedStateUpdate.related_to`, `RelatedDeltaUpdate.related_to`, `RelatedStateAndDeltaUpdate.related_to` and `RelatedContract.instance_id` were decoded with:

```rust
ContractInstanceId::from_bytes(update.related_to().data().bytes()).unwrap()
```

`from_bytes` is `bs58::decode(..)` — a base58 **string** decoder backing `FromStr` — but its doc said "Build `ContractId` from the binary representation", and the wire carries the id as 32 raw bytes (every encode site writes `key.as_bytes()`; the TypeScript SDK writes `Array.from(instance)`).

This is not "malformed input panics". Sampling 200,000 random 32-byte ids, **zero** decode successfully: a valid base58 decode needs all 32 bytes to be base58 characters, and the alphabet is 58 of 256 byte values (expected rate 2e-21). So **every** FlatBuffers UPDATE carrying a related update, and every PUT carrying a related contract, panicked the client's connection task on correct input. The PUT case stayed hidden because the loop body only runs on a non-empty vector and the TypeScript suite's fixture passes `new RelatedContractsT([])`.

### Shape 2: length-unchecked `(required)` fields (4 sites)

`DelegateKey.key`, `SecretsId.hash`, `RegisterDelegate.cipher` and `RegisterDelegate.nonce` were read into fixed-size arrays with `copy_from_slice` or `try_from(..).unwrap()`. The verifier checks that a `(required)` vector is PRESENT, not that it is the right LENGTH (`Verifiable for Vector<T>` runs `verify_vector_range` and stops), so a short field passes verification and panics the decoder. Same shape #85 fixed for `ContractKey.instance`.

`DelegateKey` is on the normal delegate path (`ApplicationMessages` and `UnregisterDelegate`), and the TypeScript SDK exports `DelegateKey` as the raw generated type with no length validation — unlike `ContractKey`, which throws a `TypeError`. Note the decoder was already half-guarded: `code_hash` one line below `key` was length-checked via `CodeHash::try_from`.

`SecretsId::try_decode_fbs` has no production caller today; it is fixed anyway so the first client to reach it does not find a panic waiting.

### Shape 3: union discriminants matched with `unreachable!()` (4 sites)

`ContractType` (PUT), `DelegateType` (RegisterDelegate), `UpdateDataType` (UPDATE) and `InboundDelegateMsgType` (ApplicationMessages). Every generated union verifier ends in `_ => Ok(())`, and `visit_union` only rejects when the type and value fields disagree on presence — with both present it hands **any** unrecognized discriminant straight to the decoder's match.

Precisely which values get there is worth stating, because the obvious guess is wrong and an earlier revision of this PR asserted it: a `NONE` discriminant does **not** reach the decoder by the ordinary route. Both the Rust and TypeScript builders elide a field equal to its default (`FlatBufferBuilder::push_slot`), so `NONE` is written as absent and `visit_union` rejects the present-value/absent-type pair first. What reaches the match is an out-of-range discriminant, or a `NONE` written explicitly by an encoder that forces defaults or is not flatc-generated. Still wire-reachable, but by a different door — and the distinction matters for the test, below.

`ContractRequestType`, `DelegateRequestType` and `ClientRequestType` already returned an error here; these four did not. All seven now share one error shape, and `ClientRequestType` reports the offending value, which it previously omitted.

### Shape 4 (found in review): the encode half of shape 1

`HostResponse`'s three related-update variants wrote `related_to.encode()` — base58 **text** — into `common.ContractInstanceId.data`. The other six `ContractInstanceId` encode sites in the same function write `key.as_bytes()`, the TypeScript SDK reads the field as a raw `Uint8Array`, and this PR's own decoder now rejects anything that is not 32 raw bytes.

It survived because Rust only *encodes* host responses and only TypeScript *decodes* them, so no Rust round-trip test ever crossed it and no TS test reads `relatedTo` back out of a notification. It is in scope because leaving it would ship a decoder that rejects what our own encoder emits, and because it falsified a load-bearing claim in this PR's own doc comment.

## Approach

Three shapes, three shared mechanisms rather than nine local patches — the point is that the next field and the next union are covered by construction:

- **`client_api::fixed_size_field::<N>(field, data)`** for every fixed-size wire field, naming the schema path so the error says `RegisterDelegate.cipher must be exactly 32 bytes; got 16 bytes` rather than "invalid data". `instance_id_from_fbs` (added by #85) now delegates to it and takes the field name, so a GET, an UPDATE and a related contract each name their own field.
- **`client_api::unknown_union_discriminant(union, d)`** for every union match.
- **`ContractInstanceId::from_bytes` renamed to `from_base58`.** The name is the root cause of shape 1 — it is what made four call sites feed a base58 parser raw wire bytes without looking wrong. `from_bytes` remains as a delegating alias so the rename is **not** a breaking change, but it is `#[deprecated]`, so it warns.

  **Heads-up for the next stdlib bump.** freenet-core calls `ContractInstanceId::from_bytes` at four sites (`server/client_api.rs:604`, `server/path_handlers.rs:385, 652, 952`) and names it in prose at three more (`server/client_api.rs:584, 1206, 1294`). All four call sites are correct uses — they parse a base58 URL path segment — but core builds with `-D warnings`, so the bump PR needs those four renamed and ideally the three doc references too. Harvest also has one call site (`common/src/lib.rs:70`, pinned to stdlib 0.6, so it only trips on a future bump). River, freenet-git and delta have none. That ripple is the intended nudge; flagging it so it is not a surprise.

## Testing

**The sweep.** One test walks all 256 discriminants of all seven unions on the decode path, asserting none panics, and that both an out-of-range discriminant and an explicit `NONE` produce **the decoder's own error message** — not merely `is_err()`. That distinction is the point: asserting only `is_err()` would pass on a verifier rejection, which reaches none of the code this PR changes, and the sweep would silently degrade to zero decoder coverage the moment a schema change made verification fail earlier. The `NONE` case builds with `force_defaults(true)` so the zero is written explicitly and genuinely reaches the match arm.

**Two source-scrape pins** enforce the conventions across every file hosting a `TryFromFbs` impl: no wire field read with a conversion that panics on the wrong length, and no union matched with `unreachable!()`. This is what actually covers the *next* decoder — the sweep's `cases` array is hand-written, so it covers a new variant on an existing union for free but not a new union.

**Per-site pins.** Each of the nine sites also has its own test, so a partial revert cannot leave the suite green — the exact failure mode #85 shipped with, where the panic fix reached three call sites and only one was pinned. The round-trip tests use an instance id containing `0x00`, `0xff` and the base58-excluded `0`/`O`/`I`/`l` characters, so a revert to the base58 decode cannot pass by luck.

**Verified by reverting, not by reasoning.** 14 mutations applied one at a time against the committed tree — every fixed site, plus the encode site, plus each source-scrape pin. Every one fails its pin. No vacuous tests.

| reverted site | test that fails |
|---|---|
| `RelatedContract.instance_id` | `put_related_contract_round_trips_the_raw_instance_id`, `put_related_contract_wrong_length_id_is_rejected` |
| `RelatedStateUpdate.related_to` | `related_state_update_round_trips_the_raw_instance_id`, `related_to_wrong_length_is_rejected` |
| `RelatedDeltaUpdate.related_to` | `related_delta_update_round_trips_the_raw_instance_id` |
| `RelatedStateAndDeltaUpdate.related_to` | `related_state_and_delta_update_round_trips_the_raw_instance_id` |
| `UpdateDataType` / `DelegateType` / `ContractType` / `InboundDelegateMsgType` `unreachable!()` | `no_union_discriminant_panics_the_decoder` (each independently) |
| `DelegateKey.key` | `delegate_key_wrong_length_is_rejected_not_panicking` |
| `SecretsId.hash` | `secrets_id_wrong_length_hash_is_rejected_not_panicking` |
| `RegisterDelegate.cipher` / `.nonce` | `register_delegate_wrong_length_cipher_or_nonce_is_rejected` |
| `HostResponse` related-update encode | `host_response_encodes_related_to_as_raw_bytes` |
| either convention pin | `no_wire_field_is_read_with_a_panicking_conversion` / `union_matches_never_use_unreachable` |

`cargo fmt`, `cargo clippy --all-targets --features net -- -D warnings`, and the full suite (85 tests, up from 75) are clean. No schema change and no regenerated-flatbuffers churn: the diff is 7 hand-written files.

## What is clean, for the record

The sweep also cleared these, so they are not silently unaddressed:

- **Stream reassembly** (`client_api/streaming.rs`) is well-defended: `total == 0`, `total > MAX_TOTAL_CHUNKS`, `index >= total`, `state.total != total` and duplicate-chunk are all checked before the indexing, so the index is provably in range and the allocation is capped.
- **The encode side has no instance of this class.** No Rust code encodes a `ClientRequest` outside tests; response encoding writes raw bytes for both key parts.
- **The container decoders do not re-transform**: both recompute the key from params+code rather than trusting a wire hash, and `DelegateKey.code_hash` is a checked pass-through.
- **The three `host_response` unions** have the same verifier fallthrough but no Rust decoder — Rust only encodes host responses.

## Review

Five independent lenses (code-first, skeptical, testing, big-picture, and a wire-format/compat lens), each blind to the others. Every finding was fixed or answered; the second commit is the response. The highest-severity one — the encode-side base58 — came from the skeptical lens and expanded the scope of this PR by one site, which is called out above rather than deferred.

Two things reviewers raised that are deliberately **not** fixed here:

- `RelatedContracts::try_decode_fbs` pre-allocates `HashMap::with_capacity(contracts().len())` from a wire-controlled length, which is a memory-amplification vector rather than a panic. Pre-existing, a different class from this PR's scope, and its impact depends on freenet-core's frame cap. Worth its own issue.
- The TypeScript suite still builds `new RelatedContractsT([])` and has no test emitting a related UPDATE — the fixture shape that hid the PUT bug for years. The Rust builders now pin the decoder, but a captured TS-emitted PUT carrying a related contract would close it from the other side.

Refs freenet/freenet-core#4996

[AI-assisted - Claude]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,54 @@
 ## [Unreleased]
 
 ### Fixed
+- **Related-contract decoding no longer panics on well-formed requests.**
+  `RelatedStateUpdate.related_to`, `RelatedDeltaUpdate.related_to`,
+  `RelatedStateAndDeltaUpdate.related_to` and `RelatedContract.instance_id`
+  were decoded with `ContractInstanceId::from_bytes(..).unwrap()`. That
+  function is a **base58 string decoder**, and the wire carries the id as 32
+  raw bytes. This was not an edge case: a random 32-byte id essentially never
+  consists solely of base58 characters (the alphabet is 58 of 256 byte values,
+  so the odds are about 2e-21), so *every* FlatBuffers UPDATE carrying a
+  related update, and every PUT carrying a related contract, panicked the
+  client's connection task. The PUT case stayed hidden because the loop body
+  only runs on a non-empty vector and the TypeScript suite's fixture passes an
+  empty one.
+
+- **Four more length-unchecked `(required)` fields no longer panic.**
+  `DelegateKey.key`, `SecretsId.hash`, `RegisterDelegate.cipher` and
+  `RegisterDelegate.nonce` were read with `copy_from_slice` or
+  `try_from(..).unwrap()` into fixed-size arrays. The flatbuffers verifier
+  checks that a `(required)` vector is PRESENT, not that it is the right
+  LENGTH, so any client could send a short one and take down its connection
+  task. `DelegateKey` is on the normal delegate path, and the TypeScript SDK
+  exports it as the raw generated type with no length validation.
+
+- **Four union discriminants no longer hit `unreachable!()`.** `ContractType`,
+  `DelegateType`, `UpdateDataType` and `InboundDelegateMsgType` are decoded
+  with `unreachable!()` on an unrecognized discriminant, but every generated
+  union verifier ends in `_ => Ok(())`, so an unknown discriminant — including
+  `NONE` — reaches the decoder. `NONE` is not hypothetical: the TypeScript
+  SDK's `ContractContainer` and `DelegateContainer` constructors both take it
+  as their default argument. All four now return a per-request error, matching
+  what `ContractRequestType` and `DelegateRequestType` already did.
+
+  A single test now sweeps all 256 discriminants of all seven unions on the
+  decode path, so the next union added to the schema is covered without anyone
+  remembering to write a test for it.
+
+  All of the above are **FlatBuffers-only**. The native (bincode) path decodes
+  the same Rust enums directly and never reaches `try_decode_fbs`, and
+  first-party tooling uses `encodingProtocol=native` — which is why this class
+  went unnoticed and why it lands on third-party and browser clients.
+
+### Deprecated
+- **`ContractInstanceId::from_bytes` is renamed to
+  `ContractInstanceId::from_base58`.** It parses base58 *text*, not raw bytes;
+  the old name and its "build from the binary representation" doc are what
+  caused the four decode sites above to feed it raw wire bytes. `from_bytes`
+  remains as a delegating alias, so the rename is not a breaking change, but it
+  now warns. Use `ContractInstanceId::new` for raw bytes you already hold.
+
 - **`ContractKey::try_decode_fbs` no longer double-hashes the code hash.** The
   wire `code` field carries the already-computed 32-byte code hash, but the
   decoder passed it to `CodeHash::from_code`: the *hashing* constructor -
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@
   only runs on a non-empty vector and the TypeScript suite's fixture passes an
   empty one.
 
+  Stated plainly, because it is not purely a panic fix: those four fields are
+  now **raw-bytes-only**. The old decoder did accept base58 *text* there (that
+  is the one input base58 decoding handles), so a client that worked around the
+  panic by sending text at exactly those fields is rejected now. Raw bytes is
+  what the schema type carries everywhere else, including `ContractKey.instance`
+  in the same request.
+
 - **Four more length-unchecked `(required)` fields no longer panic.**
   `DelegateKey.key`, `SecretsId.hash`, `RegisterDelegate.cipher` and
   `RegisterDelegate.nonce` were read with `copy_from_slice` or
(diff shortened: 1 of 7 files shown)


sanity:
## Full-tier review: five independent lenses

Run per `~/.claude/rules/multi-model-review.md` (Full tier — wire boundary). Each reviewer read the checked-out code blind to the others; no external/Codex pass (quota-blocked, substituted with a Claude lens per the 2026-07-26 rule change).

| lens | verdict |
|---|---|
| code-first | fixes correct, would merge; findings about claims and diagnostics |
| skeptical | not clean — found the encode-side base58 bug |
| testing | coverage real, not decorative; two assertions to strengthen |
| big-picture | scope disciplined, no tests removed; CHANGELOG structure + claim accuracy |
| wire-format / compat | wire-compatible; no schema, generated-code, encode-format or bincode change |

Every finding is fixed or answered in `ba275fd`. The three that mattered:

**1. The encode half of the same bug was live (skeptical, High).** `HostResponse`'s three related-update variants wrote `related_to.encode()` — base58 text — into `common.ContractInstanceId.data`. The other six encode sites in that function write raw bytes, the TypeScript SDK reads the field as a raw `Uint8Array`, and this PR's own decoder now rejects anything that isn't 32 raw bytes. It also falsified a load-bearing claim in this branch's doc comment ("every encode site writes `key.as_bytes()`"). Fixed here rather than deferred — the alternative was shipping a decoder that rejects what our own encoder emits. Pinned by decoding the encoder's output.

**2. The `NONE` rationale was wrong, and one assertion was vacuous (code-first, skeptical, testing — independently).** All three traced the same mechanism: both builders elide a field equal to its default (`FlatBufferBuilder::push_slot`), so a `NONE` discriminant is written as *absent* and `visit_union` rejects the present-value/absent-type pair before any decoder runs. The "TypeScript SDK constructors default to NONE" justification was false — it appeared in five comments, the CHANGELOG and the PR body — and the sweep's NONE assertion passed purely on the verifier's error, so it would have stayed green with `unreachable!()` restored. The sweep now builds that case with `force_defaults(true)`, which genuinely reaches the match arm.

Worth recording that the *out-of-range* half of the sweep was doing the real work throughout, so the original mutation results held; the NONE assertion was decorative, not wrong.

**3. The sweep could rot silently (testing, Medium).** It asserted only `is_err()`, which cannot distinguish a decoder rejection from a verifier rejection. It now asserts the decoder's exact message for both cases, so a change that makes verification fail earlier turns the sweep red instead of silently emptying it.

Also fixed: the "covers the next union automatically" claim was false (the cases array is hand-written) — two source-scrape pins now enforce both conventions across every file hosting a `TryFromFbs` impl, which is what actually covers the next decoder; three union arms that still hand-rolled their error text (`ClientRequestType` omitted the offending value entirely); `DelegateKey.code_hash` still returning `"invalid data"` one line below the field that got an actionable message; error field names naming the containing table rather than the leaf vector; a guessed `since = "0.8.5"` against a 0.8.4 manifest; two sentences of stdlib-internal rationale being shipped to clients in the wire error; one of three related variants' field names unpinned; and three CHANGELOG problems (a new heading re-parenting #85's entries, #85's now-false "still panics" note, and the raw-bytes-only behaviour change going unstated).

Mutation testing re-run after the fixes: **14 mutations, every one fails its pin**, including the encode site and both source-scrape pins.

Deliberately not fixed, both raised and both out of this PR's class:
- `RelatedContracts::try_decode_fbs` pre-allocates a `HashMap` from a wire-controlled length (memory amplification, not a panic; impact depends on core's frame cap). Deserves its own issue.
- The TypeScript suite still has no test emitting a related UPDATE or a non-empty `RelatedContractsT` — the fixture shape that hid the PUT bug. The Rust builders pin the decoder; closing it from the TS side is follow-up work.

[AI-assisted - Claude]

Issue #106, opened 2026-08-31
Thread a request id through contract responses so same-key requests can be correlated


sanity:
## Problem

Contract responses carry no request identifier, so a client cannot tell which of its own requests a response belongs to. The only identifying field on a `ContractResponse` is the contract key (or, for `NotFound`, the instance id).

That is enough to stop one contract's answer reaching another contract's caller — [#94](https://github.com/freenet/freenet-stdlib/pull/94) does exactly this, and it fixed the real user-visible harm. It is **not** enough for two requests on the *same* key, which are byte-identical on the wire.

The open case, reported in #96:

1. `get(K)` is sent and exceeds the client's `REQUEST_TIMEOUT_MS`. The SDK rejects the promise, but sends nothing to cancel the operation — the node keeps working on it.
2. The caller retries `get(K)`.
3. The first request's answer arrives and matches the retry's key exactly, so it settles the retry. The retry's own answer is later dropped against an empty queue.

Usually this means staler state for the correct contract. Not always: a retry issued with `fetchContract: true` can be settled by an earlier response carrying no contract.

## Why this cannot be fixed client-side

A fence was attempted in [#105](https://github.com/freenet/freenet-stdlib/pull/105) and withdrawn. The approach — record a debt when a request is abandoned, and let the next response for that key settle the debt instead of a live request — fails for an information-theoretic reason, not an implementation one:

**The SDK cannot tell the ghost's answer from the retry's own answer.** They are the same message shape for the same key. Whichever arrives first is absorbed. And the likely ordering is the *unhelpful* one: the abandoned request timed out precisely because it was slow, while the retry may be served quickly.

So the fence drops the retry's own answer, the retry hangs until its own timeout, and that mints another indistinguishable case. The chain sustains itself for as long as the caller retries, producing spurious timeouts against a contract the node is serving correctly — strictly worse than the mis-delivery it removes. An expiry window does not bound it, because each debt is spent long before it can expire.

Two further facts make any "the answer will still arrive" assumption unsafe:

- freenet-core's session actor delivers **at most one `HostResponse` per (transaction, client)** — `handle_result_delivery` removes the transaction entry on first delivery. A host `Error` *is* the operation's terminal response, so a fence that assumes a success can still follow it records a claim that will never be paid.
- An operation can die without emitting a client result at all (the result-router channel dropping under saturation, `pending_results` LRU eviction). "Eventually answered" is not guaranteed.

## Proposed fix

[...]

PR #110, opened 2026-09-01
build: pin flatc and check flatbuffers binding drift in CI

sanity:
Prep for #106. No schema, wire format, or hand-written code changes — this makes the *next* change (adding a `request_id` field) a small reviewable diff instead of an unreviewable one.

## Problem

The flatbuffers bindings are checked in, but `rust/build.rs` shelled out to whatever `flatc` was on `PATH` and rewrote them in place **on every build** — and did nothing at all when there was no `flatc` to find. Both halves cause real problems.

**Churn.** Different flatc releases format their output differently, so a contributor whose compiler didn't match got thousands of lines of unrelated diff on every build, whether or not they'd touched a schema. Not hypothetical: commit `8070e93`, *"revert: undo accidental flatbuffers regen from #72"*, exists because it reached `main`.

**The silent fallback is worse.** CI doesn't install flatc, so **CI never regenerated anything** — it compiled the checked-in bindings and reported success. A schema edit that was never regenerated would pass CI while the bindings on disk still described the *previous* schema. Nothing anywhere compared the two.

The bindings had in fact drifted from every flatc release available: they matched neither 2.0.8, 24.3.25, nor 25.9.23 output, even after formatting.

## Approach

- Regeneration is **explicit** (`FREENET_REGEN_FLATBUFFERS=1`) and refuses to run under a flatc other than the pinned one.
- A new CI job `flatbuffers_drift` regenerates with the pinned compiler and fails if the result differs from what's committed. **That job, not the build, is what enforces the schema↔bindings relationship.**
- The job verifies the flatc version before comparing, because a silently-different compiler would make it compare the wrong thing and pass — the exact failure mode it exists to prevent.

## Why 24.3.25

Chosen on evidence, not preference:

- It matches the `flatbuffers = "24.3"` crate version already in `rust/Cargo.toml`.
- Regenerating the **TypeScript** bindings with it is a **no-op** against what's committed — so it's already the de-facto compiler for that half.
- The Rust bindings it produces pass the **byte-frozen wire-format tests unchanged**, which is the evidence that the wire format is untouched and only generated-code style moves.

## Commits

1. **`0cbeca8`** — the mechanism (`build.rs` + CI job). This is the commit with the argument in it.
2. **`45523cc`** — the regenerated output. Mechanical; doesn't need line-by-line reading.

## Testing

- Regeneration is **idempotent** — identical diff on a second run, so the drift check is stable.
- A plain `cargo build` leaves a deliberately-reverted binding file reverted, confirming the guard works in both directions.
- 56 Rust tests pass, **including all three byte-frozen wire-format pins**; clippy clean; 78 TypeScript tests pass.

## Note for reviewers

There is a related gap this PR does **not** close: the request side has three byte-frozen wire-format pin tests, but the **response side has none** — nothing currently fails CI if `HostResponse`'s bincode layout moves. Given #106 is specifically about steering around that break class, it should get a pin. Kept separate to keep this PR mechanical.

Refs #106

[AI-assisted - Claude]


sanity (commented):
## Review summary: #110

**Tier:** Full (CI / build configuration is a high-risk surface).
**Lenses:** `skeptical` ×2 (separate heads) + **Codex** ×3 as the external non-Claude pass.
**Rounds:** four. Every round found something real, all of it in code added by this PR.
**Final HEAD:** `f64109a1`, CI green including `flatbuffers_drift`.

### Fixed

| Sev | Finding |
|---|---|
| **High** | `rerun-if-changed` declared `../schemas/...` unconditionally. Those paths do not exist in a registry checkout, and cargo treats a missing one as permanently dirty — so freenet-stdlib and its whole reverse-dependency cone recompiled on **every** build (~1s per no-op for this crate alone). A regression introduced by this PR. |
| **High** | The pin's justification was false: `flatbuffers = "24.3"` is a caret range resolving to **24.12.23**, and `Cargo.lock` is untracked. Re-pinned to 24.12.23. The two compilers emit byte-identical output for these schemas (verified independently by the reviewer), so no generated file changed. |
| **Med** | `build.rs` silently succeeded when explicit regeneration could not run — worse than the code it replaced, since cargo caches a *successful* build-script run. Now panics. |
| **Med** | The drift check used `git diff --exit-code`, which ignores untracked files — so a new `.fbs` generating brand-new bindings, the loudest possible drift, sailed past it. Now `git status --porcelain`. |
| **Med** | The stale-schema warning fired on **every fresh clone** (git writes `rust/src/generated/` before `schemas/`, so schemas are always strictly newer). It emitted 15 warnings across 5 jobs in this PR's own green run — pre-desensitised before it ever meant anything. Replaced with an FNV-1a content fingerprint. Verified: touching a schema warns nothing, editing one warns; **0 warnings in CI**, down from 15. |
| **Med** | Orphaned bindings were invisible: when a table moves namespace, flatc stops emitting its old file but does not delete it, so it is never modified and reads as clean. `typescript/src/host-response/application-message.ts` was exactly this and sat green through several runs. Removed, and the TS side now gets the same delete-before-regenerate treatment as Rust, keyed on flatc's own header comment. |
| **Med** | Watching only individual schema files gave cargo no input that changes when a schema is **added**. Now watches the directory too. |
| **Med** | The schema list was a hardcoded array of three while the TS side globbed, so a new schema would get TS bindings and no Rust ones — with the drift job reporting success over a file set that excluded it. Now discovered. |
| Low | Drift-failure advice still named the old pin after re-pinning, so following it installed a compiler the build script rejects. Now derived from the pin. |
| Low | MSRV: `is_none_or` is 1.82, crate is 1.80. Caught by CI clippy, not by me. |
| Low | flatc download unchecksummed; version duplicated in CI; `npm install` unnecessary; a comment about cargo's default file scan that the reviewer empirically disproved; TS glob `*` vs Rust `.fbs`; unreachable error message under `bash -e`. |

### Not fixed, with reasons

- **rustfmt is not pinned while flatc is.** `build.rs` shells to `cargo fmt` after regenerating, so a contributor on a different toolchain could produce formatting the drift job rejects while the message blames flatc. The fix is a repo-wide `rust-toolchain.toml` affecting every contributor and job — a larger decision than this PR should make alone. Worth doing deliberately.
- **`main` requires no status checks**, so `flatbuffers_drift` reports drift rather than preventing its merge. Adding it to branch protection is a repo-settings change, noted in `build.rs` rather than asserted away.
- **The bindings lose `unsafe_op_in_unsafe_fn` hygiene** by pinning to a 24.x compiler, which will need revisiting before an edition-2024 migration. Public item set is identical to `main` across all three files (reviewer-verified), so no API break.

### Verified, not asserted

Regeneration is idempotent; a plain build leaves a reverted binding file reverted; deleting all 71 TypeScript and all 3 Rust generated files and regenerating returns a byte-clean tree; two consecutive builds cache both in-repo and with `schemas/` moved away; the wrong-compiler panic fires with both versions named; 56 Rust tests, 78 TypeScript tests, clippy clean.

**Verdict: Ready to merge.**

[AI-assisted - Claude]

PR #128, opened 2026-09-07
feat(delegate): report whether a subscribe actually pinned the contract

sanity:
## Problem

`DelegateCtx::subscribe_contract` answers a three-state question with a `bool`.

`true` means **"the node accepted the registration"**, not "the contract is pinned". A registration accepted without registering durable demand — the node holds no state for the contract, or a pin cap was hit — is reported identically to one that pinned. The delegate then believes it holds durable interest and does not: the contract is not exempt from eviction, so notifications stop arriving at a moment the delegate never observes.

From the consumer case in freenet-core#5565: *a seller whose 51st order silently is not pinned misses the payment notification, and that failure is indistinguishable from "the buyer has not paid yet."* Silent under-delivery of a payment notification is money.

The `bool` is lossy twice over — it also collapses every negative host error code into `false`, so a transient failure is indistinguishable from an unknown contract.

## This is deliberately NOT a wire-format change

freenet-core#5565 states that fixing this "needs a wire change" because `SubscribeContractResponse.result` is `Result<(), String>` with no third outcome. **That is true of the V1 message path and not true of the fix**, because the V2 path is already a host function whose return channel is an `i64`:

```rust
fn __frnt__delegate__subscribe_contract(id_ptr: i64, id_len: i32) -> i64;
```

The information channel is already wide enough. What discards it is the wrapper, `result == 0` into a `bool`.

So this adds a host function and a plain Rust type. It consumes **no bincode variant tag** and cannot shift one. That matters for the release this belongs to: it does not compete with freenet-stdlib#82 or #98 for a tag, and it cannot be the thing that breaks a deployed delegate.

## Why a host function rather than a new enum variant

Both were available. The host function has the strictly better failure mode, and the difference is not stylistic:

| | old peer → new peer | new peer → old peer |
|---|---|---|
| **new enum variant** | fine, existing tags unchanged | **hard bincode decode error**, mid-protocol, unpreventable |
| **new host function** | delegate that does not import it is unaffected | fails to **instantiate**, named missing-import, at load time |

A new inbound variant is only safe if the host sends it *exclusively* in reply to something an older delegate cannot have sent — a property of the host implementation, not of the wire format, and therefore something a future change can silently violate. A host import cannot be gotten wrong that way: name resolution at instantiation either succeeds or fails loudly, before the delegate has touched any state.

This is the same argument #98 makes for `list_subscriptions`, applied to the same surface.

## `list_subscriptions` does NOT subsume this

freenet-core#5565 offers two options and prefers option 2, a delegate-visible pin-state query, as "subsuming" option 1. **It does not**, and the two should not be closed as duplicates of each other.

freenet-stdlib#98's `DelegateCtx::list_subscriptions` reads freenet-core's `DELEGATE_SUBSCRIPTIONS` (`crates/core/src/wasm_runtime/native_api.rs:41-43`). A registration that was **accepted but not pinned is in that map** — it is inserted unconditionally once the contract's code resolves, with no pin check. So a delegate that subscribed, was not pinned, and then asked what it is subscribed to would be told it *is* subscribed. The same over-claim, one call later.

Introspection answers **"what did I register?"**. This PR answers **"what actually got pinned?"**. Only the second is the question #5565 asks, and no amount of reading the subscription set can reconstruct it, because the distinction is not recorded there.

They compose: `list_subscriptions` tells a delegate what it holds after a restart; `subscribe_contract_checked` tells it whether a given registration means anything durable. Neither replaces the other.

## Approach

- `DelegateCtx::subscribe_contract_checked(&[u8; 32]) -> Result<SubscribeOutcome, i32>`
- `SubscribeOutcome` — `Pinned`, `NotPinned`, `Unrecognized(i64)`, `#[non_exhaustive]`.

`Err` means the subscription did not happen at all. A call that succeeded but did not pin is `Ok(SubscribeOutcome::NotPinned)`, **not** an error — reusing `Err` for it is what the issue rules out, because delegates legitimately subscribe before the node settles and a usually-transient condition surfacing as a hard failure would break River today.

`is_pinned()` is true **only** for `Pinned`. A newer node may report an outcome added after a given delegate was compiled; folding that into `Pinned` would reintroduce precisely the silent over-claim this type exists to remove, so it surfaces as `Unrecognized` and reads as not-pinned.

`subscribe_contract` is left **behaviourally unchanged**, deliberately. Returning a new code from it would change what already-deployed delegate WASM observes — `result == 0` would turn a soft success into a hard `false`. Its doc comment now states the loss instead.

## Verification

Run against the gate CI actually applies, plus a stricter one:

- `cargo test --features contract,net,testing,trace` — **102 passed, 0 failed** (97 before), doc-tests 1 passed
- `cargo build --target wasm32-unknown-unknown --features contract` — succeeds
- `cargo fmt --check` — clean

**The guards were made to fail before they were trusted to pass.** A test that runs can still be one that cannot fail:

| Mutation | Result |
|---|---|
| `from_code`'s unknown arm folded into `Pinned` | `an_unknown_outcome_is_not_read_as_pinned` **fails** ✓ |
| deliberate compile error inserted in the `cfg(target_family = "wasm")` branch | build **fails** ✓ — confirming that branch is genuinely compiled, not silently excluded |

The second check is here because host `cargo test` never reaches the WASM branch; #98 found twelve tests in this crate that had never executed for exactly that reason.

The off-WASM stub returns `Err(ERR_NOT_IN_PROCESS)`, never a plausible-looking outcome, and that is asserted — a stub returning `Ok(Pinned)` would let a host-side test read a pin out of a call that subscribed to nothing.

## Host half

freenet-core#5565. **No released node provides `__frnt__delegate__subscribe_contract_checked` yet**, and by the stdlib-first rule it cannot until this publishes. Core currently registers five functions in the `freenet_delegate_contracts` namespace (`crates/core/src/wasm_runtime/engine/wasmtime_engine.rs:2048-2113`); this adds a sixth.

Note for the host implementer: on current core `main`, both the V1 and V2 subscribe paths validate only that the contract's *code* is registered (`lookup_key` / `resolve_contract_key` → `code_hash_from_id`). Neither checks that state is held. So `NotPinned` is the correct answer for a contract whose code is known but which the node does not actually hold — which is the common case at node startup, not an exotic one.

## Scope

This does not address freenet-core#5566 (contract-request truncation). That issue's driving scenario does not reproduce on current `main` — see the note posted there.

[AI-assisted - Claude]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,10 @@ never arrives. On a payment address that is money. See freenet-core#5565.
 node holds state for the contract and has recorded the delegate's interest.
 Under demand-driven hosting **no subscription of any kind is an absolute pin**,
 and a delegate subscription is weaker still: it registers notification interest
-only, contributes **no hosting demand**, and so does not affect eviction ordering
-at all — unlike a client subscription, which is a ranking dimension. The name
+only. On freenet-core as it stands (pre-#4669) it contributes **no hosting
+demand**, and so does not affect eviction ordering at all — unlike a client
+subscription, which is a ranking dimension. freenet-core#5493 implements #4669
+and is open now, so that is current behaviour rather than a fixed property. The name
 describes a live subscription, not a retained one. `NotPinned` is correspondingly
 retryable, and clears as soon as the node holds the state.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,23 +2,27 @@
 
 ## [Unreleased]
 
-### Added — a delegate can learn whether its subscribe actually pinned
+### Added — a delegate can learn whether its subscribe has anything to fire on
 
-`DelegateCtx::subscribe_contract`'s doc now describes at length that a subscribe
-may register no demand, and that **a delegate cannot detect which it got**. This
-adds the way to detect it.
+`true` from `subscribe_contract` means "the node accepted the registration", and
+says nothing about whether the node holds the contract. A node that knows the
+contract's code but holds no state for it registers the interest and returns
+`true`, identically to one that holds it — which is the ordinary situation at
+startup, before the node has fetched the contract. The delegate believes it has
+a live subscription and does not, and the only signal is a notification that
+never arrives. On a payment address that is money. See freenet-core#5565.
 
-`true` from `subscribe_contract` means "the node accepted the registration", not
-"the contract is pinned". A registration accepted without durable demand is
-reported identically to one that pinned, so the delegate believes it holds
-durable interest and does not — the contract stays evictable and notifications
-later stop at a moment the delegate never observes. On a payment address that is
-money. See freenet-core#5565.
-
-- `DelegateCtx::subscribe_contract_checked(&[u8; 32]) -> Result<SubscribeOutcome, i32>`
+- `DelegateCtx::subscribe_contract_checked(&[u8; 32]) -> Result<SubscribeOutcome, i64>`
 - `SubscribeOutcome` — `Pinned`, `NotPinned`, or `Unrecognized(i64)`.
   `is_pinned()` is true only for `Pinned`: an outcome a given build cannot
-  interpret is not evidence of a pin.
+  interpret is not evidence either way.
+
+**`Pinned` is a statement about now, not a durability promise.** It means the
+node holds state for the contract and has recorded the delegate's interest.
+Under demand-driven hosting **no subscription of any kind is an absolute pin** —
+a subscribed contract is ordered last for eviction, not exempt from it — so the
+name describes a live subscription, not a retained one. `NotPinned` is
+correspondingly retryable, and it clears as soon as the node holds the state.
 
 `subscribe_contract` is left **behaviourally unchanged**, deliberately. Altering
 what it returns would change the behaviour of already-deployed delegate WASM.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,45 @@
 
 ## [Unreleased]
 
+### Added — a delegate can learn whether its subscribe actually pinned
+
+`DelegateCtx::subscribe_contract`'s doc now describes at length that a subscribe
+may register no demand, and that **a delegate cannot detect which it got**. This
+adds the way to detect it.
+
+`true` from `subscribe_contract` means "the node accepted the registration", not
+"the contract is pinned". A registration accepted without durable demand is
+reported identically to one that pinned, so the delegate believes it holds
+durable interest and does not — the contract stays evictable and notifications
+later stop at a moment the delegate never observes. On a payment address that is
+money. See freenet-core#5565.
+
+- `DelegateCtx::subscribe_contract_checked(&[u8; 32]) -> Result<SubscribeOutcome, i32>`
+- `SubscribeOutcome` — `Pinned`, `NotPinned`, or `Unrecognized(i64)`.
+  `is_pinned()` is true only for `Pinned`: an outcome a given build cannot
+  interpret is not evidence of a pin.
+
+`subscribe_contract` is left **behaviourally unchanged**, deliberately. Altering
+what it returns would change the behaviour of already-deployed delegate WASM.
+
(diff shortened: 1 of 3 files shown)


sanity:
## H1 is fixed — current head is `969a9a6`, not `31db23b4`

The blocker was addressed one commit after the prelude/`i64` fixes. Quoting the pushed content rather than asserting it, since the finding was verified against the earlier SHA:

```rust
pub enum SubscribeOutcome {
    /// **The node holds state for this contract, and has recorded the
    /// delegate's interest in it.** Notifications have something to fire on.
    ///
    /// This is a statement about *now*, not a durability promise. Under
    /// demand-driven hosting **no subscription of any kind is an absolute
    /// pin** — a subscribed contract is ordered last for eviction, not exempt
    /// from it — so a delegate must not read this as "the node will keep this
    /// contract for me". It means the subscription is not vacuous today.
    Pinned,
```

"enters the renewal set" and "exempt from eviction" are gone. `NotPinned` is now "registered, but the node holds **no state** for the contract", and says explicitly that it clears once the node holds the state — so it is retryable against a condition that actually changes, which was the substance of the blocker. `CODE_CAP_EXCEEDED` deliberately not added.

I verified the three supporting claims against freenet-core before editing, rather than taking them on report:

- `DELEGATE_SUBSCRIPTIONS` appears **nowhere** in `ring/hosting` — inserted on subscribe, read for notification fan-out, cleaned up on unregister, feeding no hosting decision.
- `docs/design/demand-driven-hosting.md` mentions delegates **zero times** (`grep -ci delegate` → 0) and says of the subscriber dimensions "neither is an absolute pin" (`:200`).
- `ring/hosting/cache.rs:899` — a subscribed contract is "NOT hard-pinned", corroborated at `:30`, `:1643`, `:1947`, `:6118`.

### One thing that looks like a second instance and is not

`delegate_host.rs:760` still contains "the renewal set and is not exempt from eviction". That line is **#98's, already on `main`**, and it states the **negative** — that nodes predating freenet-core#4669 register no demand, so the contract does *not* enter the renewal set and is *not* exempt. It describes what a node lacks rather than promising a pin, so it is accurate and deliberately untouched.

### Also fixed here

- **M7** — `SubscribeOutcome` re-exported from `prelude`.
- **M8** — `Result<_, i64>`, matching `list_subscriptions` and the host channel; the `unwrap_or(i32::MIN)` clamp is gone.

The prelude guard is worth a note. Its first version used `use crate::prelude::*` inside a test module that already has `use super::*` in scope, so the name resolved from the parent and **the test passed with the re-export deleted** — confirmed by deleting it. It now names `crate::prelude::SubscribeOutcome` in full; deleting the re-export fails the build with `E0433`. The reason is in the test's doc comment so it is not "simplified" back.

131 tests pass, both clippy invocations clean, fmt clean, CI green on `969a9a6`.

[AI-assisted - Claude]

sanity:

**This bypasses a misconfigured branch protection, not a review or a CI gate.** `freenet-stdlib` main uses classic branch protection with `required_pull_request_reviews` present and `required_approving_review_count: 0` — a configuration that blocks every merge while requiring nothing that can satisfy it. The rules API reports all enforced rules passing, which is why the block is confusing to diagnose. Recent stdlib merges have gone through the `Freenet Developers` bypass allowance for the same reason. The config is worth fixing.

What the bypass is **not** skipping:

- **All 8 required checks green** at `59c8dc13739d7f429059406c61bbc7a5b15f3076`, the exact commit merged.
- **Four independent review passes**, final verdict clean. The blocking finding — that `Pinned` was defined as "enters the renewal set and is exempt from eviction", neither of which has a referent in core — was fixed by adopting the observable meaning the PR already carried elsewhere. `DELEGATE_SUBSCRIPTIONS` feeds nothing in `ring/hosting`, and `contract_in_use` (`ring/hosting/reconcile.rs:543-547`) is `has_local_client || has_farther_downstream_subscriber || has_recent_local_client_access` — three terms, none a delegate.
- Both new guards verified by mutation: deleting the prelude re-export now fails with `E0433`, and folding the unknown arm into `Pinned` fails `an_unknown_outcome_is_not_read_as_pinned`.

[AI-assisted - Claude]

PR #130, opened 2026-09-07
docs: move the silent-loss claim down a level, to where it is true

sanity:
Follow-up to #129, which merged while this was being written. Documentation only.

## What was wrong

#129's "Resolving a conflict in a test module" section has the mechanism, the structure and the guidance right. The **instance** is wrong — and the correct one arises from the identical cause and is strictly worse.

It attaches the silent-loss claim to a lost `#[cfg(test)]` on a module. That is not a silent-loss mode: every test still runs, because `#[test]` functions are collected by the harness regardless of it, and `cargo build` then warns about unused imports because the test code compiles into the library. Losing it makes the build **noisier**. (#129 already says this correctly in its table; this PR makes the surrounding prose lead with the case that earns the claim.)

## The instance that does earn it

Same mechanism, one level down. Reproduced in a scratch repo rather than reasoned about — two branches each appending a `#[test]` function to the **same** module:

```
     3	    #[test]
     4	    fn base() {}
     5	    #[test]          <- leading context: shared, belongs to neither side
     6	<<<<<<< HEAD
     7	    fn tb() {}
     8	=======
     9	    fn ta() {}
    10	>>>>>>> sideA
    11	}                    <- trailing context: also shared
```

Keep both sides and you get `#[test] fn tb() {}` followed by a **bare** `fn ta() {}`. Braces balance. It compiles. It runs. `ta` is quietly no longer a test, and **the suite is green with one fewer test in it.**

## Leading versus trailing, which is the actual point

#129 calls both the attribute and the brace "shared *trailing* context". Only the brace is trailing; the attribute is **leading**, and that is the dangerous one:

- a lost **trailing** brace is a hard compile error — loud, immediate, unmissable;
- a lost **leading** attribute is not an error at all, because what remains is still valid code meaning something else.

Same mechanism, opposite consequence, and only the harmless one announces itself. The section now leads with the attribute for that reason.

## One refinement from measuring rather than assuming

A lost `#[test]` does emit a `dead_code` warning, so it is not *literally* without signal. What makes it dangerous is that the warning lands in build output nobody reads on a green run, while the signal that matters — the test result — stays green and merely gets smaller. Only a function absorbed wholesale into a neighbour has no possible signal, since no referent is left to warn about. The table says so.

## Unchanged

The census guidance is untouched. It was right while the mechanism under it was wrong, and it is exactly what sees this: losing `ta`'s attribute drops that module's count by one, which a by-name-and-count check catches and nothing else does.

## Verification

- `cargo test --features contract,net,testing,trace` — 146 passed, 0 failed (unchanged; this PR touches no code)
- `cargo fmt --check` — clean

Files: `WIRE-FORMAT.md`, `CHANGELOG.md`.

[AI-assisted - Claude]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,13 +24,13 @@
   value only the host can act on.
 
   Also records that resolving a conflict in a **test module** by keeping both
-  sides is safe for the content and not for the delimiters — a closing brace or
-  an attribute can sit in shared context and be dropped — with a measured table
-  of what the compiler does and does not catch. Losing `#[cfg(test)]` is *not* a
-  silent-loss mode: every test still runs and the build gets noisier. The
-  dangerous cases are a dropped `#[test]` (absent from the run, green with a
-  smaller count) and a function absorbed wholesale (no warning is possible).
-
+  sides is safe for the content and not for the delimiters. Git's conflict
+  regions are line ranges, so a `#[test]` attribute can end up in *leading*
+  shared context: two branches each appending a test to one module resolve to a
+  bare `fn`, which compiles, runs, and is quietly no longer a test — the suite
+  green with one fewer test in it. The leading case is the one that bites, since
+  a lost *trailing* brace is a hard error. Losing `#[cfg(test)]` on a module is
+  *not* a silent mode: every test still runs and the build gets noisier.
   Linked from CONTRIBUTING.md, and at the repo root rather than `docs/` because
   a bare `docs` line in `.gitignore` makes that directory untracked and
   `ci.yml`'s `paths-ignore` skips it.
--- a/WIRE-FORMAT.md
+++ b/WIRE-FORMAT.md
@@ -245,38 +245,59 @@ Two changes appending at the same anchor conflict there, and the resolution is
 almost always "keep both sides". That is right for the *content* and unsafe for
 the *delimiters*.
 
-Git's conflict regions are line ranges. They do not respect syntactic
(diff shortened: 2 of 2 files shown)


sanity:
Merging with `--admin`: this bypasses stdlib's misconfigured classic branch protection (`required_pull_request_reviews` present with `required_approving_review_count: 0` — blocks every merge, requires nothing that can satisfy it), not a review or a CI gate. All 8 checks green at `88cd949dbc68987938308323259a37aa35b259c8`. Documentation only.

Completes the conflict-hazard section merged in #129, which was correct but attached its central sentence one construct too high.

**The generalisation is the point, and it was reproduced in a scratch repo rather than reasoned about:** git's conflict regions are line ranges, so the tokens binding a construct together can land in shared context. A lost *trailing* brace is a hard compile error — loud and unmissable. A lost *leading* attribute is not an error at all, because what remains is still valid code meaning something else. Same mechanism, opposite consequence, and only the harmless one announces itself.

So the case that bites is two branches each appending a `#[test]` function to one module: keep-both-sides yields a bare `fn`, braces balance, it compiles, it runs, and the suite is green with one fewer test in it.

One deliberate restraint worth recording: a dropped `#[test]` does emit a `dead_code` warning, so the section does not claim it is signal-free. What makes it dangerous is that the warning lands in build output nobody reads on a green run while the test result stays green and merely gets smaller. Only an absorbed function has no possible signal. Writing "nothing fails" flat would have repeated the original error in the other direction.

[AI-assisted - Claude]

